Issue #16, opened 2021-12-02
Blinded signatures revision


BasileiosKal:
As it is right now, the commitment mechanic (blinded signatures) could lead to some inconsistencies and possible forged signatures. For example, if the Issuer signs the messages `m1`, `m2`, a "standard" (non-blinded) signature would be, `(A, e, s)` with

    A = (g0 + h0 * s + h1 * m1 + h2 * m2)*(1/(SK+e))

If the Holder commits a message `m3` using `h1`, i.e., 

    commit = h0 * s’ + h1 * m3

The blinded signature, created by the Issuer, would be 

    A_blind = (commit + g0 + h0 * s + h1 * m1 + h2 * m2)*(1/(SK+e)) = 
      = (g0 + h0 * (s+s’) + h1 * (m1+m3) + h2 * m2) * (1/(SK+e)). 

Notice that `(A_blind, s+s’, e)` is a valid BBS+ signature in the messages `(m1+m3)` and `m2`, which is not the desired outcome and a potential forgery (`A_blind`, `s`, `s’` and `e` are known to the Holder). 

The main problem is that currently there is no mechanism for the Issuer to validate that the Holder does not create commitments using the same G2 elements as themself (`h1` in the example above). My proposal is (at least for now) to define a blinded signature that would only contain committed from the Holder messages.

andrewwhitehead:
This is exactly why the commitment proof of knowledge needs to be verified by the issuer prior to creating a signature using the commitment.

BasileiosKal:
Thank you for your answer. I had misunderstood BlindMessagesProofGen. I think we should also mention that the Issuer should check that the used generators for the commitment are not also used for the signature, (i don't think it would be obvious to an implementer). I can make a pr for that and i think we can close this issue.

PR #17, opened 2021-12-12
Checks for the blind message generators and PoK during signing

BasileiosKal:
Related to #16.

1. Added an explicit use of BlindMessageProofVerify in BlindSign to validate the PoK of committed messages, and a check that the generators used in the commitment are not also used for the known messages.
2. Changed some of the notation to accommodate those changes above. The new notation can be also used more easily if we want to add support for blind signing with multiple commitments.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -449,18 +449,17 @@ Procedure:
 
 The PreBlindSign algorithm allows a holder of a signature to blind messages that when signed, are unknown to the signer.
 
-The algorithm takes in a generated blinding factor that is used to un-blind the signature from the signer, and a pedersen commitment from the generators in the signers public key PK and a vector of messages.
+The algorithm returns a generated blinding factor that is used to un-blind the signature from the signer, and a pedersen commitment from the generators in the signers public key PK and a vector of messages.
 
 ```
-(s', commitment) = PreBlindSign((msg[i],...,msg[U]), h0, (h[i],...,h[U]))
+(s', commitment) = PreBlindSign((msg[1],...,msg[U]), h0, CGens)
 ```
 
 Inputs:
 
-- msg\[i\],...,msg\[U\], octet strings of the messages to be blinded.
+- msg\[1\],...,msg\[U\], octet strings of the messages to be blinded.
 - h0, octet string.
-- h\[i\],...,h\[U\], octet strings of generators for the messages to
-    be blinded.
+- CGens, vector of octet strings of generators for the messages to be blinded
     
 Outputs:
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -449,18 +449,17 @@ Procedure:
...
 ```
 
 Inputs:
 
-- msg\[i\],...,msg\[U\], octet strings of the messages to be blinded.
+- msg\[1\],...,msg\[U\], octet strings of the messages to be blinded.
 - h0, octet string.
-- h\[i\],...,h\[U\], octet strings of generators for the messages to
-    be blinded.
+- CGens, vector of octet strings of generators for the messages to be blinded
```
tplooker, line 462: When we say "octet string" here are we meaning a string with some form of defined encoding like hexadecimal or instead are we assuming any parameters that are binary in nature be supplied as byte array?

spec.md
```diff
@@ -469,49 +468,60 @@ Outputs:
 
 Procedure:
 
-1. s' = H(PRF(8 \* ceil(log2(r)))) mod r
+1. (Ch[1],...,Ch[U]) = CGens
 
-2. if subgroup\_check(h0) is INVALID abort
+2. s' = H(PRF(8 \* ceil(log2(r)))) mod r
```
tplooker, line 473: Probably should define `ceil` in terminology also even though it probably obvious to most

spec.md
BasileiosKal: My understanding is that currently by octet string we mean just a byte array with no specific encoding, mainly used to represent points in extension fields using the format described [here ](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-10) in [subsection 2.5](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-10#section-2.5), or for representing nonnegative integers using the conversions described [here](https://www.rfc-editor.org/rfc/rfc8017.txt) at section 4.

tplooker:
Discussed on WG call on 15th of dec @BasileiosKal has an action to make some changes requested by @andrewwhitehead and @mikelodder7

mikelodder7 (commented):
The part about the key being 48 * (messages +1) isn’t always accurate. I would like to rename these generators to domain parameters as this is the more common with cryptography. That means the public key has a fixed size of 96 bytes. 

The reason for the distinction is the domain parameters have no secret meaning vs a public key indicates a secret key behind it

BasileiosKal:
I agree and I updated this pr.  However, it does seem that a wider update to the spec may be needed.  It may also be useful to specify the domain parameters used by each function, similarly to the format from [here](https://www.ietf.org/archive/id/draft-irtf-cfrg-rsa-blind-signatures-02.html#RemoteTiming).<br/>
It also seems like [this ](https://github.com/decentralized-identity/bbs-signature/issues/19) issue (#19 ) will be relevant i.e., if we don't change the way the generators are created, it may be confusing to refer to them as domain parameters and have them be both issuer and number of messages specific.

mikelodder7:
Agreed. Perhaps we use the current spec as an example for creating the domain parameters. The point is they need to be distinct points whose discrete log is unknown

tplooker:
Related to #19, @andrewwhitehead to supply a review to this PR

tplooker (approved):

Issue #15, opened 2021-11-23
Generalize the presentation proof nonce


christianpaquin:
The spec currently specifies a presentation nonce to prevent replay attack, and recommends it to be a random value. I suggest we had a few details and requirements around nonce generation, to accommodate various scenarios: 
 - For replay protections, the nonce only need to be unique, so a verifier identifier + timestamp would work (to create proofs non-interactively).
 - If the holder keys are held in hardware that the user only has access for a limited period of time, then making the nonce random prevents the user pre-generating proofs to be used later after the authorized period.
 - A BBS+ attribute credential can be used to create a digital signature on some data, disclosing a subset of the attributes. In this case, the nonce can be the digest of the signed data.

I can create a PR with suggested updates.

BasileiosKal:
I agree with defining requirements around nonce generation. I don’t understand though how the third option prevents replay attacks or pre-generated proofs (could not the adversary replay the presentation of the same disclosed attributes or pre-generate proofs for the whole credential??). I would appreciate if you could explain that part.

christianpaquin:
The last point is a different potential use case: digital signature on say, a document, vs. using a BBS+ credential for authn/authz. E.g., Jane is a professional engineer of company A is asked to review and sign off on the design of a project from company B. She does, and uses her work credential with attributes (name, affiliation, level, role, etc.) to sign off the document, and discloses affiliation, level, and role (e.g., "This was signed by a level 60 engineer from the Cloud group of Company A"); she does not disclose her name because her company doesn't want the customer to poach their employees. In this case, the challenge `c` is the hash of the document, just like in conventional electronic signatures. There isn't a replay attack in this case; since the attacker would simply present the same signed document with same signature and disclosed attributes.

BasileiosKal:
Thank you for the explanation. I think it’s a very interesting use-case and it would be great to have it documented in the spec. 

Maybe we should also include the proof purpose to the nonce to avoid an un-trusted verifier getting documents signed by supplying the hashes of the documents as random nonces (in an interactive protocol)??

BasileiosKal:
As a follow up to that issue, after talking with @christianpaquin I realized I had misunderstood the proposed use case. Although there could be a higher layer spec defining credentials for signing documents, I think it would be nice to also have that use case documented in the current spec and recommend a nonce accordingly (Maybe add a nonce generation section in the security considerations??).

tplooker:
Addressed in PR #55, closing on that basis

PR #21, opened 2022-01-10
Clarified presentation message requirements

christianpaquin:
Clarified usage of nonce when use in a non-interactive protocol.


tplooker:
Addresses issue #15

tplooker (approved):

tplooker:
Discussed on Jan 10th WG call, resolved to merge after 1 day if no further objection or feedback.

andrewwhitehead:
I'm not sure the text `BBS signatures are nondeterministic` is still true depending how the nonces are chosen. It also might be good to refer to `e` and `s` specifically in the context of nonce selection (at least I think those are the relevant nonces?). I'm not sure why a signed document would include a ZKP of a signature rather than the signature itself but maybe there's a use for that.

andrewwhitehead:
Hm, I also can't imagine how blinded signatures would be used in a non-interactive or deterministic context.

christianpaquin:
>I'm not sure why a signed document would include a ZKP of a signature rather than the signature itself but maybe there's a use for that.

The use case I referred [here](https://github.com/decentralized-identity/bbs-signature/issues/15#issuecomment-978239178) if you want to sign a statement without identifying yourself, think role-based signature similar to role-based access. A BBS+ cred owner could sign a statement, disclosing some attributes attested by a trusted issuer.

christianpaquin:
> I also can't imagine how blinded signatures would be used in a non-interactive or deterministic context.

One scenario that dates way back to Chaum's credential days was toll booths, where passing cars could present a payment by signing a booth ID + timestamp. Prover just broadcasts the proof to get access to a resource, using a (verifier ID + timestamp + small random value) as the nonce. Verifiers can prevent replays by keeping a local cache of used nonces (deleting as the timestamps get too old).

tplooker:
@andrewwhitehead any concrete proposal that could be suggested here to move this PR along?

christianpaquin:
Comments from the Jan 24th meeting: yes, the nonce terminology contradicts the deterministic use cases. Perhaps renaming it to something more neutral would help generalize the supported scenarios. In U-Prove, e.g., the `nonce` was called `presentation message`, which could be a nonce (for interactive scenarios), a verifierID+timestamp for user-initiated non-interactive scenarios, or static content for digital signature use cases.

spec.md
```diff
@@ -797,6 +797,12 @@ The ZKP protocols use nonces which MUST be different in each context.
 
 BBS signatures can be implemented on any pairing-friendly curve. Using BLS12-381 the signature achieves close to 128-bit security, and is the recommended curve at this time.
 
+## Nonce selection
+
+The proofs generated in this specification insures that the key material was applied on the specified nonce. A verifier-specified cryptographically random nonce provides strong protections against replay attacks. In some settings, proofs can be generated in a non-interactive fashion; in this case verifiers MUST be able to verify the uniqueness of the nonce values (provers could use a unique verifier ID combined with a timestamp to help with this).
+
+The proof generation acts a digital signature on the nonce, which could be used to design BBS+ based digital signature mechanisms. In this case, the nonce could be the hash digest of the message to sign, and the resulting proofs (potentially containing disclosed attribute values) would form the signature.
```
tmarkovski, line 804: While I share the idea that there are very good use cases where deterministic nonce can be used here, I think they make more sense in higher level protocols. Especially, since the wording used here may incorrectly guide implementors to adopt timestamp for a nonce.
I think a shorter guidance describing that this parameter be unique, and random enough would be more appropriate.

Specific implementations may still make a conscious choice to use deterministic value where security or risk impact tolerates this. For example: attendees at a conference can prepare a proof by using the room number for a conference where both verifier and presenter know this nonce and require no interaction, thus simplifying the verification experience.

spec.md
```diff
@@ -769,7 +769,7 @@ As discussed in Section 2.2, the IKM input to KeyGen MUST be infeasible to guess
    
 Secret keys MAY be generated using other methods; in this case they MUST be infeasible to guess and MUST be indistinguishable from uniformly random modulo r.
 
-BBS signatures are nondeterministic, meaning care must be taken against attacks arising from signing with bad randomness, for example, the nonce reuse attack on ECDSA [HDWH12]. It is recommended that the nonces used in signature proof generation are from a trusted source of randomness.
+BBS signatures are nondeterministic, meaning care must be taken against attacks arising from signing with bad randomness, for example, the nonce reuse attack on ECDSA [HDWH12]. It is recommended that the nonces used in signature proof generation are from a trusted source of randomness (see Nonce selection section below).
```
tmarkovski, line 772: I don't know what standard in cryptography is used to reference to good randomness. Maybe [this RFC](https://datatracker.ietf.org/doc/html/rfc4086) may be more appropriate to quote here?

tmarkovski:
> Comments from the Jan 24th meeting: yes, the nonce terminology contradicts the deterministic use cases. Perhaps renaming it to something more neutral would help generalize the supported scenarios. In U-Prove, e.g., the `nonce` was called `presentation message`, which could be a nonce (for interactive scenarios), a verifierID+timestamp for user-initiated non-interactive scenarios, or static content for digital signature use cases.

I think that's a good description of an application of a nonce, when it's used in a deterministic scenario, but not general enough. `Challenge` may be used when interaction is needed in a non-deterministic manner. A `nonce` can then become a `presentation message`, a `challenge`, or even a `pin` to be more specific. This makes me think that `nonce` is indeed a good way to refer to this input.

tplooker:
@christianpaquin apologies but we have renamed the spec file name so we leverage existing tooling around managing ID drafts, can you please rebase your PR accordingly?

christianpaquin:
I've updated the PR, merging with latest changes on main, and removing some mentions of potential use cases causing confusion.  

The main use case I have in mind for BBS+ signature is to build anonymous credentials, where the holder is not only given certified attributes, but also holds a private key (in one of the never-disclosed attributes, like in U-Prove and Idemix). Since BBS+ can, in its generalization, be used to encode key-less, bearer-only signatures on arbitrary messages, the previous "signing the nonce" text was confusing, so I removed it for now.

This being said, where should we specify such a use for BBS+ signature? Should that be a totally different spec (similar to CL sig vs. Idemix?) or a simpler profile?

spec.md
christianpaquin: Agreed that we need clearer usage "profiles" to scope the implementation choices. Really interested in defining these.

tplooker:
> This being said, where should we specify such a use for BBS+ signature? Should that be a totally different spec (similar to CL sig vs. Idemix?) or a simpler profile?

Just so I can clarify when you say a more opinionated signature scheme what do you mean exactly, one that features holder cryptographic binding or something else?

draft-bbs-signatures.md
```diff
@@ -739,6 +739,10 @@ The ZKP protocols use nonces which MUST be different in each context.
 
 BBS signatures can be implemented on any pairing-friendly curve. Using BLS12-381 the signature achieves close to 128-bit security, and is the recommended curve at this time.
 
+## Nonce selection
```
tplooker, line 742: I think somewhere we need to relate the concept of a nonce to a presentation message, is here the most appropriate place to do that?

tplooker: Given that we now no longer formally have the concept of a nonce in the draft itself

christianpaquin: > Just so I can clarify when you say a more opinionated signature scheme what do you mean exactly, one that features holder cryptographic binding or something else?

Yes, at a minimum. This could be achieved (I think) by making one of messages the "private key" and you shouldn't reveal it. I've seen in another thread someone proposing tying it to the BLS key (?), I need to look into the details of that...

christianpaquin: > I think somewhere we need to relate the concept of a nonce to a presentation message, is here the most appropriate place to do that?

Ah yes, didn't realized that changed already when I merged. I definitely need to update that section to reflect that.

christianpaquin: Updated

draft-bbs-signatures.md
```diff
@@ -742,6 +742,10 @@ The ZKP protocols use nonces which MUST be different in each context.
 
 BBS signatures can be implemented on any pairing-friendly curve. Using BLS12-381 the signature achieves close to 128-bit security, and is the recommended curve at this time.
 
+## Presentation message selection
+
+The signature proofs of knowledge generated in this specification are created using a specified presentation message. A verifier-specified cryptographically random value (e.g., a nonce) provides strong protections against replay attacks, and is RECOMMENDED in most use cases. In some settings, proofs can be generated in a non-interactive fashion, in which case verifiers MUST be able to verify the uniqueness of the presentation message values.
```
tplooker, line 747:
```suggestion
The signature proofs of knowledge generated in this specification are created using a specified presentation message. A verifier-specified cryptographically random value (e.g., a nonce) featuring in the presentation message provides strong protections against replay attacks, and is RECOMMENDED in most use cases. In some settings, proofs can be generated in a non-interactive fashion, in which case verifiers MUST be able to verify the uniqueness of the presentation message values.
```

OR13 (approved):

OR13:
how many days with approval before merge for this work item?

tplooker:
Resolved to merge on WG call 7th of Feb, awaiting conflict resolution

christianpaquin:
Conflict merge resolution is very confused on my system, so closing this and opening new, cleaner PR #55.

Issue #8, opened 2021-09-27
Crypto agility


schanzen:
The current draft is partially agnostic to the cryptographic details and in other parts very specific.
For example, the choice of curve is left open: https://github.com/schanzen/bbs-signature/blob/main/spec.md#choice-of-signature-primitive
But, the curve is implicitly enforced in https://github.com/schanzen/bbs-signature/blob/main/spec.md#terminology along with other primitives such as the hash functions.
This will make #7 a bit awkward to fix.

I guess the specification could be very specific to only define a concrete instantiation or abstract to allow crypto agility.
The latter may require additional metadata in the objects.

Related to #5 as well. Maybe hashfunctions should be harmonized for the initial recommendation/instantiation?

tplooker:
Thanks for raising, I agree along side the hashing algorithm as we discussed on the most recent call the underlying curve used should be left open. Could you elaborate on where in the terminology section this is currently not the case?

schanzen:
Under  https://github.com/schanzen/bbs-signature/blob/main/spec.md#terminology you are defining ```dst```,  ```hash_to_curve_g1```, and ```hash_to_curve_g2``` with a parameter which is tied to BLS12-381: BLS12381G1_XMD:BLAKE2B_SSWU_RO_BBS_SIGNATURES:1_0_0 and BLS12381G1_XMD:BLAKE2B_SSWU_RO, respectively.

The ```dst``` parameter should probably be different depending on the used hash and curve and so should the algo for the hash_to_curve.

schanzen:
I guess the question is if the "BBS+ Signature Scheme" is supposed look like https://datatracker.ietf.org/doc/html/rfc6979 where you then have specific instantiations in the test vectors: https://datatracker.ietf.org/doc/html/rfc6979#appendix-A.2.5
Or if it should be more like https://www.rfc-editor.org/rfc/rfc8032 where the choice of primitives is taken away (on purpose).
I guess one could make the argument that the approach of EdDSA is more sensible as implementors will make less errors, but maybe the scheme itself should then also be qualified. Ex: "The BLS12-381-SHAKE-BBS+ signature scheme" or similar.

dlongley:
> One could make the argument that the approach of EdDSA is more sensible as implementors will make less errors...

+1 to this approach.

tplooker:
[...]

Issue #7, opened 2021-09-23
Test vectors


tplooker:
In order to facilitate interoperability this specification should contain a set of test vectors for the cryptographic structures defined

tplooker:
Initial test vectors for the signature have been contributed in #67, the remaining will be contributed once the proof serialization is settled.

tmarkovski:
Will be fixed with #221

tplooker:
Adding a couple of comments here raised from #221 

- Form of messages in the test vectors is unclear, you must map these to scalar first as they are raw byte strings
- Disclosed message syntax for uses a zero based index which isn't described, other option is to use a 1 based counter as suggested by @christianpaquin

tplooker:
The residual work here is to add test vectors for the proof protocol.

Issue #2, opened 2021-09-13
Document the binary formats for cryptographic structures defined by the specification


tplooker:
Within the draft there are several new cryptographic structures that are encoded in binary, their format should be documented to support interoperable implementations, these include

- ProofOfKnowledge (signature proof)
- Commitment
- Commitment ProofOfKnowledge
- BBS Signature

tplooker:
@andrewwhitehead

tplooker:
A topic for future discussion is whether to use a structured (e.g CBOR) vs custom binary format, current implementations use a custom binary format however structured solutions like CBOR could be considered for future revisions.

ChristopherA:
I’m hoping you document CBOR versions of these (and not the IPFS version of CBOR). We are increasingly using CBOR in wallet interoperability standards as it allows airgap solutions like UR: 

* https://github.com/BlockchainCommons/crypto-commons#urs
* https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md

tplooker:
#66 adds a preliminary serialisation format for the signature, however the proof representation is still un-documented

tplooker:
Signatures have been addressed in #66 only outstanding rules are for proofs therefore closing this issue in favour of #153

PR #22, opened 2022-01-10
Add various comments as FIXMEs

schanzen:
Related to #7 and #8

Did a pass of the document and was wondering to what degree it is supposed to be normative. If it is **not** supposed to be, it should be made clear(er) what the diff over the respective papers (informative references to those missing) is.
Also, normative references (such as those to the HKDF) are not required then.

If it is supposed to be normative, it requires the definition of (at least) one concrete instantiation (with hash functions/curves) and normative test vectors and wire formats.

Once we can clarify those points I can replace the FIXMEs aligning with the target.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -52,20 +52,19 @@ A digital signature scheme is a fundamental cryptographic primitive that is used
 
 However traditional digital signatures are limited to fixed mode of signing and verifying, that is the entire payload that was signed by a signer must be known by the verifier in-order to validate the digital signature.
 
-This document describes the BBS signature scheme. The scheme features many important properties:
+This document describes the BBS signature scheme. (FIXME: THE BBS signature scheme? Or A BBS signature schemes? Also: BBS citation needed/informative reference).
+The scheme feature important properties that allow the scheme to be used in applications where privacy and data minimization techniques are desired and/or required:
 
-1. The signature is over a group of Pedersen commitments--signatures can be created blinded or un-blinded.
+1. Signatures can be created blinded or un-blinded.
    
-2. The signature is encoded as a single group element and two field elements.
+2. Signatures are encoded as a single group element and two field elements. (FIXME: Why is this important? Succinctness? Unclear)
 
-3. Verification requires 2 pairing operations.
+3. Verification requires 2 pairing operations. (FIXME: Only two? Why is this important? It's not as good as one or no pairing!)
 
-4. Simple signature schemes require the entire signature and message be disclosed during verification. BBS allows a fast and small zero-knowledge signature proof of knowledge to be created from the signature and the public key. This allows the signature holder to selectively reveal any number of signed messages to another entity (none, all, or any number in between).   
+4. Traditional signature schemes require the entire signature and message to be disclosed during verification. BBS allows a fast and small zero-knowledge signature proof of knowledge to be created from the signature and the public key. This allows the signature holder to selectively reveal any number of signed messages to another entity (none, all, or any number in between).   
     
-These properties allow the scheme to be used in applications where privacy and data minimization techniques are desired and/or required.
-     
 A recent emerging use case applies signature schemes in [verifiable credentials](https://www.w3.org/TR/vc-data-model/). One problem with
-using simple signature schemes like ECSDA or ED25519 is a holder must disclose the entire signed message and signature for verification. Circuit based logic can be applied to verify these in zero-knowledge like SNARKS or Bulletproofs with R1CS but tend to be complicated. BBS on the other hand adds, to verifiable credentials or any other application, the ability to do very efficient zero-knowledge proofs. A holder gains the ability to choose which claims to reveal to a relying party without the need for any additional complicated logic.
+using simple signature schemes like ECDSA or ED25519 is that a holder must disclose the entire signed message and signature for verification. Circuit based logic can be applied to verify these in zero-knowledge like SNARKS or Bulletproofs with R1CS but tend to be complicated. BBS on the other hand adds, to verifiable credentials or any other application, the ability to do very efficient zero-knowledge proofs. A holder gains the ability to choose which claims to reveal to a relying party without the need for any additional complicated logic. (FIXME: Informative references missing)
 
 ## Notational Conventions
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -836,6 +842,8 @@ The following comparison assumes BBS signatures with curve BLS12-381, targeting
 
 For 128 bits security, ECDSA with curve P-256 takes 37 and 79 micro-seconds to sign and verify signature on a modern computer. BBS 680 and 1400 milliseconds to sign and verify a single message. However, ECDSA can only sign a single message whereas BBS can sign any number of messages at the expense of a bigger public key. To sign and verify 10 messages takes 3.7 and 5.4 milliseconds, and 22.3 and 24.4 milliseconds for 100 messages.
 
+(FIXME: How is sign and verify of 10 messages faster than a single message? Also, it seems to scale linearly, as would an ECDSA signature over 10/100 messages)
```
andrewwhitehead, line 845: I think those 680 and 1400 numbers are also in microseconds not milliseconds?

spec.md
```diff
@@ -52,20 +52,19 @@ A digital signature scheme is a fundamental cryptographic primitive that is used
...
 
-This document describes the BBS signature scheme. The scheme features many important properties:
+This document describes the BBS signature scheme. (FIXME: THE BBS signature scheme? Or A BBS signature schemes? Also: BBS citation needed/informative reference).
+The scheme feature important properties that allow the scheme to be used in applications where privacy and data minimization techniques are desired and/or required:
 
-1. The signature is over a group of Pedersen commitments--signatures can be created blinded or un-blinded.
+1. Signatures can be created blinded or un-blinded.
    
-2. The signature is encoded as a single group element and two field elements.
+2. Signatures are encoded as a single group element and two field elements. (FIXME: Why is this important? Succinctness? Unclear)
```
andrewwhitehead, line 60: Why does it need justification?

spec.md
```diff
@@ -52,20 +52,19 @@ A digital signature scheme is a fundamental cryptographic primitive that is used
...
 
-3. Verification requires 2 pairing operations.
+3. Verification requires 2 pairing operations. (FIXME: Only two? Why is this important? It's not as good as one or no pairing!)
```
andrewwhitehead, line 62: Same here, not sure what the suggested change would be.

spec.md
schanzen: Because above it it says "The scheme features many important properties:"
It is unclear why this generic property is important.
I changed this sentence above in the pull request already.
If this is just a general note then it is not an important property of the scheme I think.

spec.md
schanzen: It also kind of implies that there is space time tradeoff and that BBS can sign any number of messages in 680/1400ms. So maybe the 3.7 and 5.4 actually refer to ECDSA. Not sure.

spec.md
schanzen: Well let me rephrase it using the words from the (original) text:
How is "Verification requires 2 pairing operations." a "[property that] allow[s] the scheme to be used in applications where privacy and data minimization techniques are desired and/or required."?
That is what the text currently says, and it is unclear (to me) where the connection is.

spec.md
schanzen: Also, the "encoding" of the signature would usually refer to a normative (interoperable) wire format of the signature. But the encoding  Maybe what is meant here is that the signatures are "represented" as a single group element and two field elements?
Later in the text it is also stated that "How a signature is to be encoded is not covered by this document. (TODO perhaps add some additional information in the appendix)"

spec.md
```diff
@@ -690,6 +695,7 @@ Procedure:
 28. return spk
 
 How a signature is to be encoded is not covered by this document. (TODO perhaps add some additional information in the appendix)
+(FIXME: Encoding out of scope OK, but wire format is required for test vectors. Unless this document is not normative.)
```
tplooker, line 698: I think I agree with your comment here but could you clarify what you mean by "wire format" specifically and what you would take as the definition of encoding format?

spec.md
```diff
@@ -262,6 +264,8 @@ Procedure:
 SkToDpk algorithm takes a secret key SK and outputs a corresponding short formed public key.
 
 SK MUST be indistinguishable from uniformly random modulo r (Section 2.2) and infeasible to guess, e.g., generated using a trusted source of randomness.  KeyGen (Section 2.3) outputs SK meeting these requirements.  Other key generation approaches meeting these requirements MAY also be used; details of such methods are beyond the scope of this document.
+(FIXME: This paragraph is a duplicate from Section 2.2. where it is already established that IKM is random and cannot be guessed)
```
tplooker, line 267: If it is a duplicate then I think we should just remove it

spec.md
mikelodder7: Sure feel free to remove it.

spec.md
schanzen: As an implementer I would expect qualifiers such as "32 octets" or "variable length" (with a length field) for the members of the tuple and then a normative ordering of them (for example by having a wire format graphic or a concatenation of the octet strings).
The point is that when I eventually code my structs, I need to know the size of its members and their location.
Otherwise I do not even know how to parse a test vector (if provided).
This unclarity may have to do with #8 which makes normative statements more difficult to incorporate.

spec.md
tplooker: Yeah I think this is covered by issue #2 and we should aim to review it on the next working group call to provide better clarity around the direction the WG would like to take this

spec.md
tplooker: Yeah I think this is covered by issue #2 and we should aim to review it on the next working group call to provide better clarity around the direction the WG would like to take this

spec.md
tplooker: @schanzen, could you update your PR to just remove this section entirely from the appendix?

spec.md
tplooker: Have reviewed, @schanzen can you remove the duplicate language around SK, then we can resolve this FIX me

spec.md
tplooker: +1 reading the context it does not appear apparent why this needs to be bulleted as it is, I would vote to remove it.

spec.md
tplooker: +1 as above I think this should be removed also

andrewwhitehead (approved):

tplooker (approved):

tplooker:
Purely editorial changes, multiple approvals, discussed on WG, merging, thanks @schanzen

PR #23, opened 2022-01-10
Check the hash for the challenge of the Fiat-Shamir heuristic

### LINKED ISSUE #20: NIZK's must use fiat-shamir hash comparison
When verifying any NIZK, it is necessary to check if the fiat-shamir heuristics are the same i.e. the challenges.
Comparing whether the commitments are the same is not enough.

BasileiosKal:
Attempts to fix #20

Changes: 

1. SpkGen returns the computed challenge for Fiat-Shamir.
2. SpkVerify checks that the challenge is correct.


tplooker:
Discussed on WG call Jan 10th, waiting on a review from @mikelodder7 as the underlying issue was raised by him.

spec.md
```diff
@@ -729,21 +729,23 @@ Procedure:
...
 8. if X1 != X2 return INVALID
 
-9. c = H(Abar || A' || h0 || C1 || d || h0 || h\[i1\] || ... || h\[iR\] || C2 || nonce)
+9. c_v = H(Abar || A' || h0 || C1 || d || h0 || h\[i1\] || ... || h\[iR\] || C2 || nonce)
 
-10. T1 = Abar - d
+10. if c != c_v return INVALID
 
-11. T2 = P1 + h\[i1\] \* Rmsg\[1\] + ... + h\[iR\] \* Rmsg\[R\]
+11. T1 = Abar - d
```
mikelodder7, line 736: These steps are not necessary since they should be handled by the challenge hash.

mikelodder7: I.E 11-16 are not necessary

mikelodder7: See https://hackmd.io/Q587Q9p7T5ab30NTn4MvTA#Change-selective-disclosure-proofs-to-just-be-Schnorr-Proofs-and-Fiat-Shamir

BasileiosKal: Thank you for the link. It is very interesting. Perhaps I will have to think more about it but it seems to me that some of the steps are needed. If we change the hash calculated at step 9 here to contain Y1 and Y2 instead of C1 and C2 (and of course to be calculated in a later step), perhaps we could remove the checks at steps 15 and 16, and consequently remove C1 and C2 from the spk returned from SpkGen. However, steps 10 to 14 don't seem to me that we could remove them, they are an integral part of the Schnorr proof after all.

mikelodder7: The problem is it’s being calculated twice. Once for the hash and then again checked for equality which is why I said in my write up they can be removed

mikelodder7: If the values are different, the hash will be too.

tplooker:
@BasileiosKal apologies but we have renamed the spec file name so we leverage existing tooling around managing ID drafts, can you please rebase your PR accordingly?

OR13 (changes requested):
There is an implied change request by @mikelodder7 and there are conflicts that need to be addressed.

tplooker:
Discussed on WG call 7th Feb, awaiting review from @mikelodder7 for progress

andrewwhitehead:
It would likely be more efficient to check the challenge hash before doing the pairing operations, in case that fails.

mikelodder7:
Sure it could be. I don’t have a preference

BasileiosKal:
updated

andrewwhitehead (approved):

tplooker:
@mikelodder7 can you review these latest changes please

tplooker (approved):
Having had the chance to review this PR more closely now, I agree with this approach and how it is algorithmically described. It is consistent with @mikelodder7's original issue described in #20. The PR does appear to have some overlap with #62, however on balance I think the scope of this PR makes it easier to review and merge in first.

tplooker:
PR open for multiple weeks, discussed on WG call multiple times, multiple approvals, merging.

Issue #30, opened 2022-01-25
Naming of the draft BBS vs BBS+


tplooker:
I propose we rename the draft to "The BBS Signature Scheme" and elect to drop the + from all mentions of BBS. Through appropriate citations of the academic work I think we can still achieve the right level of association to the schemes origins. The rationale for dropping the + is that in many visible places that character is not allowed (e.g naming repositories, naming packages and programming language syntax)

quartzjer:
I definitely support this :+1:.

tplooker:
Addressed in #32 there was only two minor inconsistencies in the draft already that have been addressed

PR #32, opened 2022-01-25
Add ID tooling, remove redundant scripts

tplooker:
- Add basic tooling for managing this spec as an ID draft based on https://github.com/martinthomson/internet-draft-template.
- Removes some redundant scripts for old tooling.
- Removes old HTML version.
- Minor tweaks to draft naming based on #30


.github/update.yml
```diff
@@ -0,0 +1,36 @@
+name: "Update generated files"
+# This rule is not run automatically.
```
OR13, line 2: remove comments?

.github/setup.yml
```diff
@@ -0,0 +1,91 @@
...
+        with:
+          make: update-venue
+
+      - name: "Update GitHub Pages"
+        if: ${{ steps.first.outputs.first != 'true' }}
+        uses: martinthomson/i-d-template@v1
+        with:
+          make: gh-pages
+
+      - name: "Push Changes"
```
OR13, line 88: Looks like you are no longer using markdown2rfc.... I had to bump the version recently... https://github.com/OR13/markdown2rfc

I wanted to render latex equations as svgs...

OR13: I love the idea of github handling this all for you tho, i would still probably want to test it locally.

.github/update.yml
tplooker: It just describes how the tool is run copied directly from the template (https://github.com/martinthomson/internet-draft-template/blob/main/.github/workflows/update.yml#L2)

.github/setup.yml
tplooker: Yeah I have seen some other repos that include the ability to do this locally to preview changes in final output form, was going to tackle in a follow up PR

.github/setup.yml
```diff
@@ -0,0 +1,91 @@
+name: "Perform Initial Repository Setup"
```
tplooker, line 1: Have updated this workflow file essentially it is designed to be run once then deletes it self, just sets up tooling around the draft publication

.github/setup.yml
tplooker: Note as above this workflow is just designed to be run once also

tplooker:
Merging PR so we can get spec hosting working correctly, no substantive changes to the spec text.

Issue #19, opened 2021-12-13
More flexible default generators


andrewwhitehead:
At the moment, the default generators are defined using roughly the formula `hash-to-curve(PK || index || count)`, meaning that the generators are dependent on both the public key of the issuer and the number of signed messages.

There is also discussion of wanting to use a set of predetermined random generators instead of the default generators, in which case the generators would be independent of both the public key and the number of messages in the signature.

In the interest of aligning these approaches somewhat, I think it's worth considering removing the message count from the hash input, so that the default generators depend only on the public key of the issuer. This could have positive effects on performance given that all the generators could be cached up to a maximum number of signed messages.

This change might also be relevant in the case of JSON-LD credentials. In order to commit to one or more hidden messages (to implement holder binding for instance) a holder would need to have the relevant message generators. But the total number of messages is often not known at this point in the credential exchange protocol, because the JSON-LD credential itself (including the proof block?) would need to be generated and normalized first. If the generators were independent of the number of messages and the hidden messages were among the first messages (not added to the end) then this would be made possible.

mikelodder7:
I'm fine if we want a different formula. As long as the inputs have sufficient entropy.

tplooker:
Another option is for the message generators to be pre generated and shared globally rather than being issuer specific like the proposal above. 

Discussed on 10th Jan WG call, resolved to add a section to the spec that speaks about message generator creation which will take a seed and output a list of message generators

tplooker:
@andrewwhitehead will provide a PR that defines ^ process

andrewwhitehead:
There was some discussion on the call around a DST for the generators. This is an extra parameter to hash-to-curve (besides the seed and message index used as input) which would separate generators used for different signing algorithms. Currently the draft specifies `BLS12381G1_XMD:BLAKE2B_SSWU_RO_BBS+_SIGNATURES:1_0_0`. I think for the next draft it might be something like `BLS12381G1_XOF:SHAKE256_SSWU_RO_BBS+_SIGNATURES:2022`. This implies that the chosen seed value should not include such information, and might be something trivial like the encoding of the words `message generators` in the common case.

mikelodder7:
Message generators must NOT be shared globally. They must be unique per public key for the protocol to be secure

mikelodder7:
[...]

PR #35, opened 2022-01-26
Remove deterministic public key section from the spec

### LINKED ISSUE #26: Conceptually seperate message generators from public key definition
At the moment in the draft a BBS signature public key is defined as an ordinary public key (using the based generator for the curve group) + a set of message generators used for committing messages to a BBS signature. Given the conversation occurring in #19 even if the outcome is to stick with a set of message generators that are derived through a deterministic process from the ordinary public key, I think we should remove the concept of a BBS signature public key and regard message generators as "public parameters" instead.

tplooker:
Regardless of the outcome of #19 most in the group appear to agree that generators should always be considered an internal cryptographic structure instead of one that forces API consumers to manage externally, with that in mind we can remove any mention of deterministic public keys vs short form public keys and instead simplify to have only 1 type.

Fixes #26


tmarkovski:
`PK` has two generation paths with equivalent output. What happens with `PK` generation when `SK` isn't available, e.g. by a verifier?

tplooker:
@tmarkovski can you clarify when SK is not available what other option there is to generate PK?

tmarkovski:
I misunderstood the initial intent. You're basically suggesting we turn DPK into PK, and hide the generator construction internal the implementation, since these keys don't need to managed externally anyway.

tmarkovski (approved):

andrewwhitehead (approved):

PR #41, opened 2022-01-29
Generator demo

mikelodder7:
This is a demo of how the BBS+ message generators can be created in Rust.


tplooker (approved):

tplooker:
Thanks Mike, this is a great foundation, I'll merge this and do a small amount of re-organising so we have space for other tooling in future

PR #44, opened 2022-01-31
Add basic usecases section

tplooker:
This PR is the start of some documentation around possible usecases for the BBS scheme. Unsure about where it should live or the style these should be written in.


OR13 (approved):

draft-bbs-signatures.md
```diff
@@ -799,5 +799,34 @@ The example below illustrates the creation of a blind signature. Let the Signer
...
+
+Improved schemes around authorization protocols often involve adding a layer of proof of cryptographic key possession to the presentation of an access token, which mitigates the deficiencies highlighted above as well as providing a way to detect a replay attack. However, approaches that involve proof of cryptographic key possession such as DPoP (https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop-04) suffer from an increase in protocol complexity. A party requesting authorization must pre-generate appropriate key material, share the public portion of this with the authorization server alongside proving possession of the private portion of the key material. The authorization server must also be-able to accommodate receiving this information and validating it.
+
```
BasileiosKal, line 817: +1 for the DPoP comparison, it does seem that not much will change for the auth server though, unless we make whatever holder binding mechanism we agree on optional (which could be worth considering). IMO the biggest advantage will be when it comes to the RS, since they will only have to validate the bbs+ proof, instead of both the JWT and the DPoP.

draft-bbs-signatures.md
```diff
@@ -799,5 +799,34 @@ The example below illustrates the creation of a blind signature. Let the Signer
...
+BBS Signatures due to their unique properties removes this source of correlation but maintains the same set of guarantees required by a resource server to validate an access token back to its relevant authority. In the context of a protocol like OAuth2.0 the access token issued by the authorization server would feature a BBS Signature, however instead of the relying party providing this access token as issued, in their request to a resource server, they derive a unique proof from the original access token and include that in the request instead, thus removing this vector of correlation.
+
```
BasileiosKal, line 811: It may not be trivial how to sign a token (i.e., JSON) with bbs+. Should we point to JWP here perhaps?

draft-bbs-signatures.md
tplooker: Agree, however the nuance here is that a BBS+ signature issued even without holder binding is still better than an ordinary JWT as it gives the ability for the resource server to detect a replay attack. While keeping the protocol relatively simple as the authorization server does not need up front key material from the relying party during an authZ request, nor does the resource server need to validate both a JWT and seperate DPoP.

draft-bbs-signatures.md
tplooker: Yeah potentially a good idea as a note? Feel free to suggest one

BasileiosKal (approved):

tplooker:
No normative, informative appendix on the application of the scheme, multiple approvals, merging

PR #45, opened 2022-01-31
Fixed private key domain to 0 < SK < r

christianpaquin:
As defined in `KeyGen`, `SK` should not be `0` or `r` in `SkToPk`. The `KeyGen` definition matches selection from `Zp*` from section 4.3 of [Camenisch, Drijvers, Lehmann](https://eprint.iacr.org/2016/663.pdf).

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -265,7 +265,7 @@ PK = SkToPk(SK)
 
 Inputs:
 
-- SK, a secret integer such that 0 <= SK <= r
+- SK, a secret integer such that 0 < SK < r
 
 Outputs:
 


mikelodder7:
Probably shouldn’t be any low value either like `1`

tplooker:
@mikelodder7 can you suggest some spec text if you think this definition should be even more constrained?

tplooker (approved):

BasileiosKal:
> Probably shouldn’t be any low value either like `1`

I don't see a problem if SK is indeed random. This is similar to the way it is defined [here ](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.2) i think.

BasileiosKal (approved):

christianpaquin:
> Probably shouldn’t be any low value either like `1`

I updated the PR description to point to section 4.3 of CDL, where they pick SK at random from Zp*. This PR is mainly to correct an inconsistency in the spec.

tplooker:
Minor change, aligning to underlying academic work, multiple approvals, merging. Thanks @christianpaquin

Issue #46, opened 2022-01-31
[Proposal] An interop profile document


tplooker:
Given the discussion in #8 around how the core bbs signature draft should relate to underlying cryptographic primitive selection, I wanted to propose an approach that may be inline with this emerging consensus.

Instead of mentioning in the core bbs signature draft anything around concrete curves or digest algorithms we could define a seperate document which is an opinionated "bbs signature profile" aimed at promoting interoperable implementations. The profile would also host test vectors to facilitate this process.

If this approach is accepted by the community then I would suggest we immediately start work on one profile that uses the following.
- BLS12-381 as the pairing friendly curve
- SHAKE256 as the digest algorithm

Alternatively if having multiple document is too complex we could elect to merge these, however we would need to be careful in how we structure the document.

OR13:
I think ideally, you just name the algorithm and define it concretely for those parameters... a profile for their configuration seems a bit excessive to me.

I do think you should make it obvious how someone might replace either in your spec, but I don't think you need to take ownership of defining that generality for all of IETF.

tplooker:
Ok so a counter proposal would be to generalize section 5 https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-overview which is currently empty to be more "profiles" with the definition of BLS12-381 under that alongside concrete test vectors?

OR13:
Have a look at:

- https://www.rfc-editor.org/rfc/rfc9162.html

- https://datatracker.ietf.org/doc/html/rfc8812

for examples of registeringthings.

tplooker:
This issue has been addressed in #47 and #60, with the general consensus being around defining the BLS12-381 ciphersuite/profile within the draft

PR #47, opened 2022-02-01
Restructure to allow for profile sections

tplooker:
Minor restructure to the draft to make room for the concrete definition of a profile for the BBS signature scheme based on BLS12-381. Relevant discussion in #46

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -183,15 +183,11 @@ subgroup\_check(P) -> VALID or INVALID
 
 This document is organized as follows:
 
-* The remainder of this section defines terminology and the high-level API.
+* (#core-operations) defines core operations of the signature scheme.
 
-* Section 2 defines primitive operations used in the BLS signature scheme. These operations MUST NOT be used alone.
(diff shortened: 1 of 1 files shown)


draft-bbs-signatures.md
```diff
@@ -770,10 +762,15 @@ The cipher-suites in Section 4 are based upon the BLS12-381 pairing-friendly ell
...
 * subgroup_check MAY use either the naive check described in Section 1.3 or the optimized check given by [Bowe19].
 
-## Test Vectors
+### Test Vectors
 
 //TODO
 
+# IANA Considerations
+
+This document does not make any requests of IANA.
```
OR13, line 771: we need a spec that does update the JOSE / COSE registry... eventually. 
I would prefer if we could do that here.

OR13 (approved):

andrewwhitehead (approved):

draft-bbs-signatures.md
tplooker: See #48

tplooker:
Multiple approvals, minor re-structure of doc, blocking further progression of draft, merging.

PR #53, opened 2022-02-05
Ensuring SK > 0

BasileiosKal:
Update `keyGen` to always return SK>0 and `KeyValidate` to check if SK = 0. 

Another option would be to point to the [BLS draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.3) since the functions are the same. There is useful context for their choices [here](https://github.com/cfrg/draft-irtf-cfrg-bls-signature/issues/25).

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -222,14 +222,21 @@ Definitions:
 - "BBS-SIG-KEYGEN-SALT-" is an ASCII string comprising 20 octets.
 
 Procedure:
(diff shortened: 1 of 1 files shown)


mikelodder7 (approved):

tplooker (approved):

tplooker:
Discussed on Feb 7th WG call, no objections, aligning to changes in underlying spec that the section was born from, merging

PR #61, opened 2022-02-14
Add message mapping to scalar

mikelodder7:
Adds a section describing how to map messages to a scalar that can be cryptographically signed.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -76,10 +76,10 @@ D
 msg
 : An input message to be signed by the signature scheme.
 
-h\[i\]
+H\[i\]
 : The generator corresponding to a given msg.
 
-h0
+H0
 : A generator for the blinding value in the signature.
 
 signature
@@ -137,11 +137,11 @@ P1, P2
 : points on G1 and G2 respectively. For a pairing-friendly curve, this document denotes operations in E1 and E2 in additive notation, i.e., P + Q denotes point addition and x \* P denotes scalar multiplication. Operations in GT are written in multiplicative notation, i.e., a \* b is field multiplication.
 
 hash\_to\_curve\_g1(ostr) -> P
-: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest
+: The cryptographic hash function that takes an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest
 algorithm, once selected apply the isogeny simplified SWU map to compute a point in G1 using the random oracle method. The domain separation tag value is dst.
 
 hash\_to\_curve\_g2(ostr) -> P
-: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G2 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G2 using the random oracle method. The domain separation tag value is dst.
+: The cryptographic hash function that takes an arbitrary octet string input and returns a point in G2 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G2 using the random oracle method. The domain separation tag value is dst.
 
 point\_to\_octets(P) -> ostr
 : returns the canonical representation of the point P as an octet string. This operation is also known as serialization.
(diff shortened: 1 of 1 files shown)


draft-bbs-signatures.md
```diff
@@ -245,6 +245,27 @@ Procedure:
 8. return SK
 ```
 
+## MapToScalar
+
+MapToScalar algorithm takes an arbitrary value and maps it to a scalar value. 
+
+Data can be mapped multiple ways depending on how it will be used with other Zero-Knowledge Proofs. For example, it is not useful to hash an integer to a scalar using SHA256 and expect it to work properly with Range Proofs.
+The following is a list of mappings and use cases:
```
tplooker, line 253: Think we should talk about usecases elsewhere

```suggestion
The following is a list of mappings:
```

draft-bbs-signatures.md
```diff
@@ -245,6 +245,27 @@ Procedure:
 8. return SK
 ```
 
+## MapToScalar
+
+MapToScalar algorithm takes an arbitrary value and maps it to a scalar value. 
+
+Data can be mapped multiple ways depending on how it will be used with other Zero-Knowledge Proofs. For example, it is not useful to hash an integer to a scalar using SHA256 and expect it to work properly with Range Proofs.
```
tplooker, line 252:
```suggestion
Data to be signed can be mapped multiple ways depending on how it will be used downstream in various Zero-Knowledge Proof protocols. For example to be able to construct a range proof for a signed message, it must have been encoded in a particular way during the signing procedure.
```

draft-bbs-signatures.md
```diff
@@ -245,6 +245,27 @@ Procedure:
...
+The following is a list of mappings and use cases:
+
+1. **Hashed**: for arbitrary length data fields that will not fit in the base field e.g. images, strings, biometrics, blockchain transaction info. 
+   1. Use [hash\_to\_field](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#name-hash_to_field-implementatio) with the domain separation tag `BBS_PLUS_BLS12381FQ_XOF:SHAKE-256_`.
+2. **Bytes**: a value that is already in the base field. No mapping metho required.
```
tplooker, line 257:
```suggestion
2. **Bytes**: a value that is already in the base field. No mapping method required.
```

draft-bbs-signatures.md
```diff
@@ -245,6 +245,27 @@ Procedure:
...
+3. **Numbers**: for range proofs. The value is zero-centered by take computing the base field modulus `z`=`q` / 3 integer division as the zero center and adding the positive number or substracting the negative number. To support complex numbers like decimal, the number is converted to fixed point arithmetic. The [Q format](https://en.wikipedia.org/wiki/Q_(number_format)) is used for these numbers as Q64.160. This leaves two bytes to avoid numbers greater than `q`. While it does not represent the full breadth of IEEE754 numbers, it does give a considerable resolution -2<sup>63</sup> to 2<sup>63</sup>-2<sup>-160</sup> signed and 0 to 2<sup>64</sup>-2<sup>-160</sup> unsigned or about 48 decimal places of precision. If decimals are rounded to the nearest integer, then Q format is not necessary.
+4. **Null**: Hard coded as 5. Means the value is not included or not used.
```
tplooker, line 259: Can you explain the assignment of the value 5?

mikelodder7: Its the smallest safe prime. Otherwise there's no other special meaning.

tplooker:
Discussed briefly on WG call 21st of Feb, one proposal was to have the MapToScalar as a pre-sign pre-spkgen operations for converting messages into their appropriate scalars.

quartzjer (approved):
Very supportive of adding these flexible definitions of different mapping methods.

quartzjer:
@mikelodder7 Is MapToScalar the same extension point where membership values would be encoded such that set (non-)membership proofs can be later generated?

draft-bbs-signatures.md
```diff
@@ -137,11 +137,11 @@ P1, P2
...
 
 hash\_to\_curve\_g2(ostr) -> P
-: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G2 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G2 using the random oracle method. The domain separation tag value is dst.
+: The cryptographic hash function that takes an arbitrary octet string input and returns a point in G2 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G2 using the random oracle method. The domain separation tag value is dst.
 
 point\_to\_octets\_min(P) -> ostr
 : returns the canonical representation of the point P as an octet string in compressed form. This operation is also known as serialization. (FIXME: This either requires a normative wire format or a reference to the normative wire format)
```
OR13, line 147: consider adding an issue for the FIXME, so can be tracked during issue review,

tplooker: @OR13 this issue has already been addressed and updated in the latest draft see https://identity.foundation/bbs-signature/draft-bbs-signatures.html#section-1.2 PR is just out of date

mikelodder7:
> @mikelodder7 Is MapToScalar the same extension point where membership values would be encoded such that set (non-)membership proofs can be later generated?

Yes

tplooker:
I understand the use case for the first three encoding methods suggested here, however unsure what value the remaining few provide, could we remove these or elaborate more on their usage.

I also think this section needs to be restructured formally to be more like the other operations defined in the spec, as currently the numeric encoding options for example describes the algorithm in a paragraph instead of say how the sign operation is performed.

Also we have to be very confident in the collision resistant properties of the numeric encoding method to prevent a message forgery attack like what is being discussed in #74.

And guidance needs to be added for the second option around message generator selection/validation which is effectively a committed message, because failure to do so could lead a signer to use a generator twice.

tplooker:
I'd also note that option 2 is not actually a MapToScalar function as the input is already a scalar so I think we can disregard it from this section of the spec and instead supply appropriate guidance elsewhere.

quartzjer:
> I understand the use case for the first three encoding methods suggested here, however unsure what value the remaining few provide, could we remove these or elaborate more on their usage.

I believe these fixed values are intended as a means for the holder to generate a proof about a value that isn't being disclosed.  If the issuer sets the value to null/empty/ignored, that the holder can generate a proof of such to differentiate that message from one that is simply being hidden.

While a similar signaling is also possible at the application level by having encoded values, I believe these are useful to represent when the value itself (encoded or not) is in the given null/empty/ignored state when being signed.  The alternative is to consider that an error, that every message MUST always have a defined value before it can be mapped to a scalar.

tplooker:
> I believe these fixed values are intended as a means for the holder to generate a proof about a value that isn't being disclosed. If the issuer sets the value to null/empty/ignored, that the holder can generate a proof of such to differentiate that message from one that is simply being hidden.

Thanks @quartzjer makes sense, in that case I understand the potential for having a representation for null or ignored (maybe?). But empty appears to be better addressed at the application layer because not all data types have the concept of empty (e.g numeric data types).

I continue to be cautious here, in the case of the numeric mapping function it must be as collision resistant as a hash function to be acceptable.

mikelodder7:
It doesn’t have to be collision resistant to map to the scalar. My first name will always map to the same value as a null would to another. They don’t even have to be unique. However mapping to a scalar should be a separate discussion in an appendix as guidelines not requirements.

mikelodder7:
I’m going to update to be such

OR13 (approved):

tplooker:
I'd like to hold off merging this PR while we discuss a proposal to break each mapping method out like is shown in #77

tplooker:
Discussed on the WG call, going to close this PR in favour of opening seperate ones for each mapping method, if not objection from @mikelodder7

PR #62, opened 2022-02-14
Update algorithms for consistency and simplicity

mikelodder7:
This updates `Sign` to be deterministic and includes updating SpkGen and SpkVerify to be mathematically correct and removes superfluous checks in verify and makes it simpler.
SpkGen proof is much smaller now.
SpkVerify has reduced steps but its equally correct and secure.


BasileiosKal:
For the `spkGen` and `spkVerify` are those essentially the same changes as in PR#23 with different notation?? 

I can close PR#23 since it seems to do the same thing for those functions.

mikelodder7:
No there not exactly the same

draft-bbs-signatures.md
```diff
@@ -318,19 +320,25 @@ Outputs:
...
+4. e = OS2IP(k) mod q. If e = 0, go back to step 3.
 
-5. A = b * (1 / (SK + e))
+5. r = h.read(64)
 
-6. signature = (point_to_octets_min(A), e, s)
+6. s = OS2IP(r) mod q. If s = 0, go back to step 5.
 
-7. return signature
+7. b = P1 + H0 * s + H[i] * msg[i] + ... + H[n] * msg[L]
```
andrewwhitehead, line 335: `b = P1 + H0 * s + H[1] * msg[1] + ... + H[L] * msg[L]` I think, and probably `msg[1]` instead of `msg[i]` above. Should we capitalize B?

BasileiosKal:
> No there not exactly the same

Ok. Is there a reason to keep both PRs open then??

It does seem that every change in #23 is also here.

draft-bbs-signatures.md
```diff
@@ -392,57 +400,57 @@ Procedure:
...
+13. r3 = r1 ^ -1 mod q
 
-14. for i in RIdxs m\~\[i\] = H(PRF(8\*ceil(log2(r)))) mod r
+14. for i in RIdxs m~[i] = HASH(PRF(8*ceil(log2(q)))) mod q
 
-15. A' = A \* r1
+15. A' = A * r1
 
-16. Abar = A' \* -e + b \* r1
+16. Abar = A' * -e - B * r1
```
andrewwhitehead, line 431: `+ B * r1` I believe

draft-bbs-signatures.md
```diff
@@ -137,7 +137,7 @@ GT
 e
 : G1 x G2 -> GT: a non-degenerate bilinear map.
 
-r
+q
```
tplooker, line 140: Have we changed all occurances of using r through out this draft?

draft-bbs-signatures.md
andrewwhitehead: Based on `Abar = A'k` checked by the pairing. `Abar = A'(-e) + B*r1 = A'(-e) + A'(e + k) = A'k`

draft-bbs-signatures.md
```diff
@@ -318,19 +320,25 @@ Outputs:
 
 Procedure:
 
-1. (w, h0, h) = octets_to_point(PK)
+1. (W, H0, H) = octets_to_point(PK)
 
-2. e = H(PRF(8*ceil(log2(r)))) mod r
+2. h = SHAKE256(SK  || msg[i] || ... || msg[L])
```
tplooker, line 325: Assume this should actually be generalized here?
```suggestion
2. h = HASH(SK  || msg[i] || ... || msg[L])
```

draft-bbs-signatures.md
```diff
@@ -318,19 +320,25 @@ Outputs:
...
 
-3. s = H(PRF(8*ceil(log2(r)))) mod r
+3. k = h.read(64)
```
tplooker, line 327: Need to document this .read() notation somewhere

tplooker (changes requested):
See comments above

tplooker:
In general im supportive of the direction this PR is taking so thanks @mikelodder7 for submitting it, however it is difficult to review because it is attempting to do many things, including:

- Updates the sign procedure to be deterministic
- Changing SpkGen procedure, including the resulting serialization of the proof
- Changing SpkVerify procedure, including how it relies on the proof as its input (now uses the challenge supplied)
- Multiple updates to notation including changing terms from r to q? Captializing some notation h -> H And using new notation that isn't completely defined e.g `hash.read()`

@mikelodder7 have I got the above right as the scope of this PR ^ and if so any way we can break it apart for independent review of each item?

draft-bbs-signatures.md
```diff
@@ -475,33 +483,25 @@ Procedure:
...
+8. cv = HASH(PK || Abar || A' || D || C1 || C2 || pm)
 
-13. Y2 = d \* r3^ + h0 \* s^ + h\[i1\] \* m^\[i1\] + ... + h\[iR\] \* m^\[iR\] - T2 \* c
+9. if c != cv return INVALID
 
-14. if C1 != Y1 return INVALID
+10. if A' == 1 return INVALID
 
-15. if C2 != Y2 return INVALID
+11. if e(A', W) * e(Abar, -P2) != 1 return INVALID
```
BasileiosKal, line 502: Why not just check for `e(A', W) == e(Abar, P2)`??

mikelodder7:
> > No there not exactly the same
> 
> Ok. Is there a reason to keep both PRs open then??
> 
> It does seem that every change in #23 is also here.

It’s not exactly. That PR has the dual commitment equality check that isn’t needed

tplooker:
@mikelodder7 can you address the reviews above and conflicts after merging #23?

draft-bbs-signatures.md
mikelodder7: Because mathematically it’s quicker. You can compute what I put with a single pairing versus two pairings. Most folks don’t know that. It’s also preferable this way.

mikelodder7:
> @mikelodder7 can you address the reviews above and conflicts after merging #23?

Planning on in the next day or so

draft-bbs-signatures.md
BasileiosKal: Got it. Maye be we should replace it with `e(A' + Abar, W - P2) != 1` then??

draft-bbs-signatures.md
mikelodder7: It matches all the other changes I've made.

draft-bbs-signatures.md
mikelodder7: No its different.

draft-bbs-signatures.md
mikelodder7: Well its can be written that way but if you were to actually compute it that way it might not work.

BasileiosKal: If that's the case then i don't see the difference with checking for equality. Don't you still compute 2 pairings??

andrewwhitehead:
I can't get the proof to work out with these changes so far, is there an implementation to compare against?

draft-bbs-signatures.md
tplooker: So you are introducing a dependency on a concrete hash algorithm here? I think previous conversations in the working group and documented in issues like #8 #46 #5 was to have this abstracted away?

draft-bbs-signatures.md
mikelodder7: I can give you a code sample to illustrate

BasileiosKal: Please do!! thanks!!

draft-bbs-signatures.md
```diff
@@ -514,31 +524,25 @@ Procedure:
...
+4. (W, H0, H[1],...,H[L]) = PK
 
-9. Y2 = d * r3^ + h0 * s^ + h[i1] * m^[i1] + ... + h[iR] * m^[iR] - T2 * c
+5. C1 = (Abar - D) * c + A' * e^ + H0 * r2^ 
 
-10. c_v = H(Abar || A' || h0 || Y1 || d || h0 || h[i1] || ... || h[iR] || Y2 || nonce)
+6. T = P1 + H[i1] * Rmsg[i1] + ... H[iR] * Rmsg[R]
 
-11. if c != c_v return INVALID
+7. C2 = T * c + D * (-r3^) + H0 * s^ + H[i1] * m^[i1] + ... + H[iR] * m^[iR]
```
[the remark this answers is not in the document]
mikelodder7: Yes it holds. I have code to prove it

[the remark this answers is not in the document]
BasileiosKal: Great! I would be interested seen the code. Regardless I would also be more at ease if we could prove correctness with math since this PR does not use exactly the formulas from the academic paper (i.e., `s’ = s + r2*r2` instead of `s’ = s - r2*r2` which is used in the paper etc.). But I may be missing something and the code could help.

draft-bbs-signatures.md
```diff
@@ -395,57 +405,57 @@ Procedure:
...
 
-23. r2^ = r2~ - c * r2
+23. r2^ = r2~ + c * r2
 
 24. r3^ = r3~ + c * r3
 
-25. s^ = s~ - c * s'
+25. s^ = s~ + c * s'
 
 26. for i in RIdxs: m^[i] = m~[i] - c * msg[i]
```
andrewwhitehead, line 456:
```suggestion
26. for i in RIdxs: m^[i] = m~[i] + c * msg[i]
```

tplooker:
@mikelodder7 can you update this PR to resolve the conflict?

tplooker:
Discussed on WG call 21st of Feb, unable to progress at this time, PR conflicts need to be resolved

BasileiosKal (approved):

andrewwhitehead (approved):

tplooker:
Merging this PR on the understanding that @BasileiosKal and @mikelodder7 are working behind the scenes to ensure that this description of SpkGen is still consistent with academic literature.

PR #63, opened 2022-02-15
Add basic section on subgroup selection for the scheme

### LINKED ISSUE #59: Document rationale for why BBS Signatures are only efficient when the signatures generated are in G1 and public keys are in G2
Unlikes BLS signatures which as per [the spec](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-02#section-2.1) define two possible variations in the scheme based on which sub-group is used for the signatures and public keys. Because of the unique mathamatical computations involved for proof generation (SPK gen) being only efficient when performed in G1, the effective variant of BBS signatures that would instead use G2 is impractical. We should document this in the spec.

tplooker:
Fixes #59


andrewwhitehead (approved):

tplooker:
Multiple approvals, purely editorial, merging

Issue #54, opened 2022-02-05
Indexes and messages notation


BasileiosKal:
The spec makes heavy use of lists of messages and indexes. We need a consistent way to represent those.

**- Option 1**:
Represent indexes with a consisted format like `I_t`, `I_d` corresponding to the total and disclosed messages. Represent the messages with something like,

    (msg[i] for i in I_d)   // or I_t

As such `spkGen` for example will be,

    spkGen(spk, PK, I_t, I_d, (msg[i] for i in I_d), pm)

**- Option 2**:
Define that all signatures on `L` messages will use the `L` first generators. With that, instead of the indexes of all the signed messages we can pass just the `totalSignedMessages`. That will also allow for a solution like [here](https://github.com/decentralized-identity/bbs-signature/issues/10#issuecomment-1018041832) with

    spkGen(spk, PK, revealedMessages, totalSignedMessages, pm)

where `revealedMessages` is a map between index and message. The drawback here is that when creating commitments for blind signatures, the user will need to choose the `n` first generators, with `n = Total_Committed_Messages` (i.e., always follow the example flow from the [blind signatures spec](https://github.com/decentralized-identity/bbs-signature/blob/main/draft-blind-bbs-signatures.md)), which could be restrictive for some applications.

**- Option 3**:
A compination of both, i.e.,

    spkGen(spk, PK, Indexes, revealedMessages, pm)

where again `revealedMessages` is a map between index and message and `Indexes` the indexes of all the messages

tplooker:
With the merge of #64 is there anything outstanding with this issue @BasileiosKal? or can we close this?

BasileiosKal:
Yeap. Further discussions on generators and indexes is better to be held elsewhere (like #68). Closing.

PR #64, opened 2022-02-16
Messages and generators notation update for readability

BasileiosKal:
First attempt to close issue #54 (see in the issue for other options).

This PR proposes to pass the generators directly to functions instead of their indexes, with the goal of increasing spec readability.

IMO this simplifies notation. For example, using indexes in spkGen will require notation for 5 different lists of indexes. This can get confusing. By passing the generators as a list, only requires notation for 2 lists of indexes (and one less input argument in SpkGen). Also,

1. Passing the messages as a map between the message and the index of the generator will still require a lot of notation for indexes.
2. Mentioning that those generators are not necessarily the `L` first elements from the global (or not) generators list, also preserves the flexibility required from the blind signatures.
3. Mentioning that implementations may choose to pass the indexes of the generators instead and pointing to a reference implementation or perhaps a more detailed explanation in the Appendix IMO will be enough to address the efficiency of the applications concerns, while keeping the spec more readable.

Also, changes in this PR use the terminology from PR #62 to some places, but I will update it elsewhere after that PR is merged.


OR13 (commented):
Sadly, I am not qualified to review this, but it looks to be a massive improvement in documenting things.

draft-bbs-signatures.md
```diff
@@ -419,7 +433,7 @@ Procedure:
 
 13. r3 = r1 ^ -1 mod r
 
-14. for i in RIdxs m~[i] = H(PRF(8*ceil(log2(r)))) mod r
+14. for j in (j1, j2,..., jU): m~[j] = H(PRF(8*ceil(log2(r)))) mod r
```
andrewwhitehead, line 436: `m~_j` instead of `m~[j]`?

draft-bbs-signatures.md
```diff
@@ -443,9 +457,9 @@ Procedure:
 
 25. s^ = s~ - c * s'
 
-26. for i in RIdxs: m^[i] = m~[i] - c * msg[i]
+26. for j in (j1, j2,..., jU): m^[j] = m~[j] - c * msg_j
```
andrewwhitehead, line 460: `m^_j` and `m~_j`? Seems inconsistent to sometimes use the `[n]` notation for array indices.

BasileiosKal: Updated. I agree but if we move to `m_c` instead of `m^` for example then it will become `m_c_j` and `m_c_j1, m_c_j2, ...` which I don't know if I like. updated for now for consistency.

andrewwhitehead: I think we can probably keep m~ and m^ now they aren't causing formatting issues.

tplooker:
Discussed on WG call 21st of Feb, awaiting review from other WG members

tplooker (commented):
Couple of minor suggestions

tplooker:
@BasileiosKal can you please update this PR to resolve the conflicts?

andrewwhitehead (approved):

tplooker (approved):

tplooker:
Multiple approvals, PR open for 2 weeks and discussed on WG call, massive improvement in notation across the spec, merging

PR #66, opened 2022-02-18
Update signature serialisation

tplooker:
Likely not going to be the final notation used in this spec, we will probably abstract away the signature encoding and decoding rules into its own function in time, however in the meantime its important to document how the e and s components are actually sent over the wire.


andrewwhitehead (approved):

BasileiosKal (approved):

tplooker:
Minor update providing important clarification around how a signature is actually encoded or decoded, partially addresses #2, multiple approvals, merging

Issue #43, opened 2022-01-31
Supporting deterministic signatures


tplooker:
BBS Signatures as defined in academic literature are non-deterministic in nature. In general non-deterministic digital signatures can be difficult from an implementation perspective due to their dependency on a sound RNG. It causes complexities around confirming implementation correctness because the RNG often has to be mocked or seeded in a particular way. There is the potential to modify the signature construction in this draft to make signatures deterministic, @mikelodder7 could you share your proposal for this?

The main questions I think we as a WG should be asking are:
- Do we want deterministic signatures?
- Do we want to take on the scope of deterministic signatures in the current draft?
- Given we are deviating from academic literature, what security proofs do we need?

mikelodder7:
My suggestion is this:

u512() is a function that converts 64 bytes as a 512-bit number.

For regular signing:

inputs: 
- base order `q`
- secret key `x`
- messages `mi,...,mn`

1. hasher = SHAKE256(x, mi, ..., mn)
2. e = uint512(hasher.read(64)) % q
3. s = uint512(hasher.read(64)) % q
4. finish as normal

For blind signing:
inputs:
- base order `q`
- secret key `x`
- known_messages `mi,...,mn`
- commitment `U`

1. hasher = SHAKE256(x, U, mi,...,mn)
2. e = uint512(hasher.read(64)) % q
3. s = uint512(hasher.read(64)) % q
4. finish as normal

BasileiosKal:
I like this method, but a comment from [here](https://datatracker.ietf.org/doc/html/rfc6979#section-3.2) worries me, i.e., "Performing a simple modular reduction would induce biases that would be detrimental to signature security"

Is this something that we will have to also worry about?

mikelodder7:
This method already addresses the bias by pulling double the bytes. See https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-13#section-5.1.

mikelodder7:
The bias is something to be aware of, but this method handles it properly.

mikelodder7:
FYI, this is how it should be done with a single signer context. Threshold signing should not be restricted to this since all sorts of attacks can happen with deterministic signing.

tplooker:
Currently covered in #62

tplooker:
Closing with the merging of #62, resolution was to include this, initial test vectors conforming to this approach have been added in #67

Issue #38, opened 2022-01-27
Message Generators Creation Method


BasileiosKal:
Since the consensus seems to be to move to global fixed message generators shared between issuers (see #19 ) the next step would be to decide how to create those generators.  Opening this issue to create some discussion around possible proposals. My understanding is that they will be deterministically generated by a common seed (created from some nothing-up-my-sleeve value).

I would also propose for the process to be such that it will create a strong binding between index and generator (i.e., something like `hash-to-curve(seed, dst, index)`) to help with indexing in applications and avoiding ambiguities.

tplooker:
@mikelodder7 I think last time we chatted you potentially had a good proposal for how this could be done?

mikelodder7:
Here's my proposal

1. Set the global domain separator to be `GLOBAL_DST='BBS_SETUP_GENERATOR_IKM_1_0_0\0\0\0'` read as 32 bytes.
2. Decide on a nothing up my sleeve value for the GLOBAL_SEED like some Shakespeare poem or quote.
3. Hasher = SHAKE256(GLOBAL_DST, GLOBAL_SEED)

Calculate the generators as follows

4. Set the generator `DST=BBS_BLS12381G1_XOF:SHAKE-256_SSWU_RO_`
5. Set generators to empty collection
6. for i < num_generators 
7. tv = Hasher.Read(64) // Read out 64 bytes from XOF
8. generators[i] = Hash2Curve(tv, DST)
9. output generators

The advantage of this approach is if we determine the global approach to be insecure we change the GLOBAL_SEED to be the signer's public key leaving all else the same. If we change curves then DST is updated to reflect that in the name.

tplooker:
A couple of things we will need to generalise in spec text is the underlying curve and subgroup, note the [current tool](https://github.com/decentralized-identity/bbs-signature/tree/main/tooling/message-generators) is coupled to BLS12-381

mikelodder7:
Sure but the curve can easily be swapped

tplooker:
Yeap agreed, just taking a note so we don't loose track of it

BasileiosKal:
+1 for that method.

mikelodder7:
Add a check for point at infinity and the base generator.

andrewwhitehead:
Can hash-to-curve return infinity? 🤔

mikelodder7:
With very small odds 1/2^256. It’s nice to check for it in case an attacker inserts one

[...]

PR #67, opened 2022-02-18
Add basic test vectors for signature cases

tplooker:
Adds a section for test vectors associated to signature generation and verification, includes an automated script that will allow us to update the fixtures via the input JSON files over time.


quartzjer:
Uhm, I like this so much I wonder what license the TS bits are under such that I might reuse them? ;)

"Imitation is the sincerest form of flattery"

andrewwhitehead:
Looks good to me!

tplooker:
> Uhm, I like this so much I wonder what license the TS bits are under such that I might reuse them? ;)
> 
> "Imitation is the sincerest form of flattery"

Feel free! Glad it is helpful!

tplooker:
@quartzjer I have updated the tooling script to be licensed under apache 2.0 :)

tplooker:
@andrewwhitehead @mikelodder7 would you be able to test these fixtures out against your implementations for correctness?

tplooker:
A couple of notes about the current set of test fixtures:
- Using the issuer public key derived generators (because we have not merged in the algorithm described in #38 yet
- Deterministic signatures based on #43

tplooker:
I've added a note to the test vectors section that documents the experimental current state of these test vectors

quartzjer (approved):

BasileiosKal (approved):

tplooker:
Multiple approvals, test vectors are still marked as experimental, merging

PR #71, opened 2022-02-24
Document generator creation procedure

### LINKED ISSUE #38: Message Generators Creation Method

tplooker:
Initial stab at this attempting to document and resolve #38. In general I was keen to define a generic procedure that could take a seed scoped either globally, issuer or signature specific. Detail about the concrete seed we are using the BLS12-381 cipher suite is still missing.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -144,8 +144,8 @@ q
 P1, P2
 : points on G1 and G2 respectively. For a pairing-friendly curve, this document denotes operations in E1 and E2 in additive notation, i.e., P + Q denotes point addition and x \* P denotes scalar multiplication. Operations in GT are written in multiplicative notation, i.e., a \* b is field multiplication.
 
-hash\_to\_curve\_g2(ostr) -> P
-: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G2 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G2 using the random oracle method. The domain separation tag value is dst.
+hash\_to\_curve\_g1(ostr) -> P
+: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G1 using the random oracle method. The domain separation tag value is dst.
 
 point\_to\_octets(P) -> ostr
 : returns the canonical representation of the point P as an octet string. This operation is also known as serialization.
@@ -578,7 +578,7 @@ Procedure:
 
 1. h = XOF(seed)
 
-2. for i in 0 to length: generators &= hash_to_curve_g2(h.read(64), dst)
+2. for i in 0 to length: generators &= hash_to_curve_g1(h.read(64), dst)
 
 3. return generators
 ```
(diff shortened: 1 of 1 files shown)


draft-bbs-signatures.md
```diff
@@ -556,6 +555,34 @@ Procedure:
...
+
+Outputs:
+
+- generators, an array of generators
+
+Procedure:
+
+1. h = XOF(seed)
+
+2. for i in 0 to length: generators &= hash_to_curve_g1(h.read(64), dst)
```
andrewwhitehead, line 581: Maybe `generators_i =`?

draft-bbs-signatures.md
```diff
@@ -137,11 +140,7 @@ P1, P2
...
-algorithm, once selected apply the isogeny simplified SWU map to compute a point in G1 using the random oracle method. The domain separation tag value is dst.
-
-hash\_to\_curve\_g2(ostr) -> P
-: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G2 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G2 using the random oracle method. The domain separation tag value is dst.
+: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G1 using the random oracle method. The domain separation tag value is dst.
```
BasileiosKal, line 143:
```suggestion
: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm. Once selected, apply the isogeny simplified SWU map to compute a point in G1 using the random oracle method. The domain separation tag value is dst.
```

BasileiosKal: Very minor suggestion.

draft-bbs-signatures.md
```diff
@@ -137,11 +140,7 @@ P1, P2
 : points on G1 and G2 respectively. For a pairing-friendly curve, this document denotes operations in E1 and E2 in additive notation, i.e., P + Q denotes point addition and x \* P denotes scalar multiplication. Operations in GT are written in multiplicative notation, i.e., a \* b is field multiplication.
 
 hash\_to\_curve\_g1(ostr) -> P
-: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest
```
mikelodder7, line 140: I think this is too long. I would just leave it as 

```
The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve].`

File: draft-bbs-signatures.md [hunk 4]
```diff
@@ -137,11 +140,7 @@ P1, P2
...
-algorithm, once selected apply the isogeny simplified SWU map to compute a point in G1 using the random oracle method. The domain separation tag value is dst.
-
-hash\_to\_curve\_g2(ostr) -> P
-: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G2 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm, once selected the isogeny simplified SWU map to compute a point in G2 using the random oracle method. The domain separation tag value is dst.
+: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The algorithm first requires selection of the pairing friendly curve and digest algorithm. Once selected, apply the isogeny simplified SWU map to compute a point in G1 using the random oracle method. The domain separation tag value is dst.
```
tplooker, line 143:
```suggestion
: The cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve].
```

As per @mikelodder7 suggestion

draft-bbs-signatures.md
tplooker: With the merging of #64 agree

tplooker:
```suggestion
2. for i in 0 to length: generators_i = hash_to_curve_g1(h.read(64), dst)
```

mikelodder7 (approved):

BasileiosKal (approved):

tplooker:
Multiple approvals, open for two weeks, merging.

PR #72, opened 2022-02-24
Auto create message generators

tplooker:
Builds on #71 to automate re-creating these and including as test vectors in the specification.


draft-bbs-signatures.md
```diff
@@ -637,6 +664,9 @@ point\_to\_octets
 octets\_to\_point
 : follows the format documented in Appendix C section 2 of [@!I-D.irtf-cfrg-pairing-friendly-curves].
 
+dst
+: "BBS_BLS12381G1_XOF:SHAKE-256_SSWU_RO"
```
andrewwhitehead, line 668: This should have an extra underscore on the end

tplooker: @andrewwhitehead can you explain why?

andrewwhitehead: In the code it does :)

BasileiosKal (approved):

draft-bbs-signatures.md
mikelodder7: In the hash to curve spec they leave an underscore at the end as well. Its mostly for consistency.

draft-bbs-signatures.md
```diff
@@ -472,23 +471,23 @@ Let the prover be in possession of a BBS signature `(A, e, s)` with `A = B * (1/
...
+        3.  D = B * r1 + h0 * r2.
+
+  Also set,
+
         4.  r3 = r1 ^ -1 mod q
         5.  s' = s + r2 * r3.
-  
-  The values `(A', Abar, d)` will be part of the spk and are used to prove possession of a BBS signature, without revealing the signature itself. Note that; `e(A', Pk) = e(Abar, P2)` where `Pk` the signer's public key and P2 the base element in G2 (used to create the signer’s `Pk`, see [SkToPk](#sktopk)). This also serves to bind the spk to the signer's `Pk`. 
+
+  The values `(A', Abar, d)` will be part of the spk and are used to prove possession of a BBS signature, without revealing the signature itself. Note that; `e(A', Pk) = e(Abar, P2)` where `Pk` the signer's public key and P2 the base element in G2 (used to create the signer’s `Pk`, see [SkToPk](#sktopk)). This also serves to bind the spk to the signer's `Pk`.
```
mikelodder7, line 483: The values `(A', Abar, d)` are part of the spk and used to prove possession of the BBS signature, without revealing the signature itself. Note that; `e(A', Pk) = e(Abar, P2)` where `Pk` the signer's public key and P2 the base element in G2 (used to create the signer’s `Pk`, see [SkToPk](#sktopk)). This also serves to bind the spk to the signer's `Pk`.

draft-bbs-signatures.md
tplooker: Have updated this now, thanks

andrewwhitehead (approved):

tplooker:
Multiple approvals, open over two weeks, builds on #71, merging

Issue #76, opened 2022-03-06
Update core operations to clarify the form of input messages


tplooker:
Currently the core sign, verify, spkGen and spkVerify operations are silent on the form the messages take into the algorithm, based on conversations with @mikelodder7 these are most often inputed in the form of scalars as outputed by the MapToScalar function being proposed in #61. We should clarify this further in the spec.

tplooker:
Discussed on WG call 15th of August, subsequent text has clarified this significantly however supporting text at the start of the relevant operations would help to resolve any outstanding ambiguity.

PR #77, opened 2022-03-07
Initial attempt to document the MapMessageToScalar function

tplooker:
This PR is primarily designed to help support conversation around design options we have here, it is also an extension of #61 and related to #76.

Currently for the core sign, verify, spkGen and spkVerify operations take the inputed messages as a set of octet strings so we either need to.

1. Clarify that the messages supplied here can be of abitrary value and that MapMessageToScalar becomes an operation executed within sign, verify, spkGen or spkVerify.
2. OR clarify that the messages supplied here have to be valid scalars and that MapMessageToScalar MUST be called on each message prior to calling sign, verify, spkGen or spkVerify.

The complexity we are managing is not just that the messages need to be mapped to scalars its also that they could be mapped in multiple ways (e.g message -> scalar as hash or as number) so where ever this definition occurs will also have to account for this optionality (e.g sign would not only take a list of messages to sign but the mapping method for each message)

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -691,9 +691,9 @@ a function that returns the point P corresponding to the canonical representatio
 - hash\_to\_curve\_g1:
 A cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve].
 
-- hash\_to\_field: A function that follows the procedure outlined in section 5.3 of [@!I-D.irtf-cfrg-hash-to-curve]
-
 - hash\_to\_curve\_g1\_dst: Domain separation tag used in the hash\_to\_curve\_g1 operation
+
+- hash\_to\_field: A function that follows the procedure outlined in section 5.3 of [@!I-D.irtf-cfrg-hash-to-curve]
 
 - hash\_to\_field\_dst: Domain separation tag used in the hash\_to\_field operation
 
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -591,6 +591,33 @@ Procedure:
 3. return generators
 ```
 
+### MapMessageToScalar
+
+There are multiple ways in which messages can be mapped to their respective scalar values, which is their required form to be used with the sign, verify, spkGen and spkVerify operations.
+
+#### MapMessageToScalarAsHash
+
+This operation takes an arbitrary message and maps it to a scalar value, by hashing it to a point in the G2 subgroup for the target curve.
+
+```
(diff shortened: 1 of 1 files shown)


tplooker:
This PR needs to be updated to use hash_to_field not hash_to_curve

tplooker:
PR has been updated to use hash_to_field now

mikelodder7 (commented):
Remove the _g1 suffix. Scalars are not part of the g1 subgroup. They are their own field. Scalars can be used in either field to scale g1 or G2 groups.

tplooker:
> Remove the _g1 suffix. Scalars are not part of the g1 subgroup. They are their own field. Scalars can be used in either field to scale g1 or G2 groups.

Addressed

draft-bbs-signatures.md
```diff
@@ -680,6 +709,9 @@ octets\_to\_point
 hash\_to\_curve_g1
 : follows the suite defined in (#bls12-381-hash-to-curve-definition-using-shake-256) for the G1 subgroup
 
+hash\_to\_field
```
mikelodder7, line 712: This should have its own `dst` value different but similar to hash_to_curve_g1. I propose `BBS_BLS12381FQ_XOF:SHAKE-256_SSWU_RO`

mikelodder7 (commented):
LGTM except we need a DST for hash_to_field. Good job.

draft-bbs-signatures.md
tplooker: Resolved

mikelodder7 (approved):

draft-bbs-signatures.md
```diff
@@ -664,7 +691,11 @@ a function that returns the point P corresponding to the canonical representatio
 - hash\_to\_curve\_g1:
 A cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve].
 
-- dst: Domain separation tag used in the hash\_to\_curve\_g1 operation
+- hash\_to\_curve\_g1\_dst: Domain separation tag used in the hash\_to\_curve\_g1 operation
+
+- hash\_to\_field: A function that follows the procedure outlined in section 5.3 of [@!I-D.irtf-cfrg-hash-to-curve]
```
andrewwhitehead, line 696: Maybe "a cryptographic hash function" as with hash_to_curve?

tplooker: Cool done

andrewwhitehead (approved):

tplooker:
Multiple reviews and approvals, PR open a week, no objections, merging

Issue #80, opened 2022-03-08
Define the hash_to_curve operation using SHAKE-256 XOF for BLS12-381 ciphersuite


tplooker:
Currently the [latest hash to curve draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14) only defines [suites](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14#section-8.8) for the BLS12-381. However due to the efficiencies involved in using SHAKE-256 we need to define the suite within this spec.

BasileiosKal:
Not really clear to me what the efficiencies are. SHAKE-256 doesn't seem to have a clear advantage over SHA256 from [here](http://bench.cr.yp.to/results-hash.html#amd64-skylake) and [here](https://keccak.team/sw_performance.html).  @mikelodder7 is the intent to use the faster KangarooTwelve(which is not standardized if I'm not mistaken) or the ParallelHashXOF256?? Also why not use ParallelHash256 if thats the case (do we need an XOF for hash2curve??).

mikelodder7:
We’re using shake in other areas of the code and it is faster when used in hash to curve. Hash to curve requires multiple calls to sha256 vs one to shake

BasileiosKal:
That clears it up! Thanks! I'm curious as to why since that is the case, the spec opts for use of expand_message_xmd with sha256 instead of expand_message_xof with shake256 when it comes to bls12-381.

tplooker:
> That clears it up! Thanks! That helped a lot to clarify it! I'm curious as to why since that is the case, the spec opts for use of expand_message_xmd with sha256 instead of expand_message_xof with shake256 when it comes to bls12-381.

Likewise, this was not immediately clear to me either especially considering other hash_to_curve suites elect to use SHAKE-256 defined in the draft.

tplooker:
Addressed in #81

PR #81, opened 2022-03-08
Add hash to curve definition for BLS12-381 SHAKE-256 suite

tplooker:
As described in #80 currently the [latest hash to curve draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14#section-8.8) does not define a suite for BLS12-381 which uses the extendable output function of SHAKE-256 used by this draft in the CreateGenerators operation defined in #71 and the MapMessageToScalar operation defined in #61 for the [BLS12-381 ciphersuite](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-bls12-381-ciphersuite), this PR establishes that definition so we can point to it.

Note I would like to review and merge this PR before proceeding with #71 #61

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -119,7 +119,7 @@ a || b
 I \ J
 : For sets I and J, denotes the difference of the two sets i.e., all the elements of I that do not appear in J, in the same order as they were in I.
 
-\[n\]: Denotes all integers from 1 to n. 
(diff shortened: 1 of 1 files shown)


mikelodder7 (approved):

BasileiosKal (approved):

tplooker:
Multiple approvals, discussed in issue, merging as it is blocking PR #71 and #61

PR #84, opened 2022-03-15
Add definition for HASH function

tplooker:
We currently use a HASH function separate to how we use an XOF function in the KeyGen SpkGen and SpkVerify functions, this provides an updated definition resolving H -> HASH instead of having two notations to describe the same thing AND provides a concrete definition for the BLS12-381 cipher suite including how many bytes we draw for the operations.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -653,7 +653,7 @@ This section defines the format for a BLS ciphersuite. It also gives concrete ci
 
 ## Ciphersuite Format
 
-- H: a cryptographic hash function.
+- HASH: a cryptographic hash function.
 
 - point\_to\_octets:
 a function that returns the canonical representation of the point P as an octet string.
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -671,7 +671,7 @@ A cryptographic hash function that takes as an arbitrary octet string input and 
 ## BLS12-381 Ciphersuite
 
 HASH
-: SHAKE-256 as defined in [@!SHA3] where the length of bytes read is ceil(log2(q)) bits, where q is the order of the subgroups G1 and G2 defined by the pairing-friendly elliptic curve
+: SHAKE-256 as defined in [@!SHA3] where the length of bytes drawn from it is dependent on the operation (32 bytes when used in the context of the KeyGen operation and 64 bytes when used in the context of SpkGen)
 
 XOF
 : SHAKE-256 as defined in [@!SHA3]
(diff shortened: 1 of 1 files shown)


tplooker:
Have updated this PR to change direction, instead of having both a HASH function and XOF this PR collapses into just using the XOF function.

mikelodder7 (commented):
A couple of notes

- Let’s remove hkdf for the secret key generation since the Xof serves the same purpose and simplifies things.
- There’s no advantage to hashing the salt first. We should just do H = XOF(salt, ikm), okm= H.read(64), sk = okm mod q, if sk=0 read more bytes and reduce again, output sk
- The challenge hash needs to draw 64 bytes and be converted to a scalar to complete the proving protocol. Verify should do the same thing. The reason is the challenge is used to compute the schnorr proofs for the hidden messages.

tplooker:
Will do, thanks @mikelodder7

BasileiosKal:
Should we be drawing different number of bytes depending on the EC used since we mod after?? For example, for [bn462](https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-10.html#name-bn-curves-for-the-128-bit-s), should we draw at least 74 bytes for 128 bits of security?? (using [this](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-14.html#name-security-considerations) method that @mikelodder7 pointed out and/or lemma 6 from [here](https://eprint.iacr.org/2009/050.pdf))

tplooker:
> Should we be drawing different number of bytes depending on the EC used since we mod after?? For example, for [bn462](https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-10.html#name-bn-curves-for-the-128-bit-s), should we draw at least 74 bytes for 128 bits of security?? (using [this](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-14.html#name-security-considerations) method that @mikelodder7 pointed out and/or lemma 6 from [here](https://eprint.iacr.org/2009/050.pdf))

Thanks, @BasileiosKal lets parameterise this in the cipher suite section, can you clarify the need to draw 74 bytes applies when using the BN 462 curve for key gen? Or also the challenge hash?

tplooker:
I have opened #87 so we can look at what this would look like

tplooker:
> The challenge hash needs to draw 64 bytes and be converted to a scalar to complete the proving protocol. Verify should do the same thing. The reason is the challenge is used to compute the schnorr proofs for the hidden messages.

Ok I will file a seperate PR to make this change

tplooker:
As discussed on the last WG call, we have decided to stick with using the HASH function but define how to use an XOF as the HASH function when defining a concrete cipher suite

PR #86, opened 2022-03-18
BBS spk security against the issuer

### LINKED ISSUE #75: BBS spk security against an untrusted/compromised issuer
Let’s say that the issuer gets compromised, i.e., both their secret key `x` and the records with the signatures they signed will become known. Let’s assume a holder has a bbs+ signature `(A, e, s)` from that issuer with `A = B * (1/(e+x))` where `B = P1 + H0*s + H_1*msg_1 + … + H_L*msg_L`. Any spk will be of the form, `spk = (A', Abar, D, ...)` where 
```
A' = A * r1 =  B * r1 * (1/(e+x))
```
 and 
```
Abar = A' * x =  B * r1 * (x/(e+x))
```
The values `(B, B*x, A', Abar)` essentially form a DDH tuple, which is easy to solve using pairings. Specifically, it will hold, 
```
e(B*x, A’) = e(B*x, B*r1*(1/(e+x))) = e(B, B) ^(r1*(x/(e+x))) = e(B, B*r1*(x/(e+x)))
```
meaning that,
```
e(B*x, A') = e(B, Abar)
```
So, if an adversary who knows `x` and `s`, sees (or has already seen) a proof with `A’` and `Abar`, they can just try different messages, construct `B` and check the above equality between the pairings. Since the messages will not be many this is not hard to do and will result in the secret messages of the holder been revealed. 

If the Issuer is untrusted, they can use the above equation to track the holder by correlating the spk's with specific signatures.

This is mitigated a little by using blind or bound signatures. If the commitment is not revealed an adversary will not be able to construct `B` by trying possible hidden messages. However, it is reasonable to assume here that if the issuers secret key and the signatures they signed are revealed so will the commitments (they will be known -and possibly recorded- by the issuer). Furthermore, if the issuer is untrusted. they can still correlate the holder.

As a result, the bbs+ proofs are not secure against an untrusted or compromised issuer. This is not that much of an issue and is like what happens in most PKIs but it might be worth documenting it in the security considerations in the spec so we will have an accurate security model.

BasileiosKal:
Closes #75 

Adding a section to the security considerations to document the fact that a bbs spk is not zero knowledge against the issuer,

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -647,6 +647,12 @@ The ZKP protocols use nonces which MUST be different in each context.
 
 BBS signatures can be implemented on any pairing-friendly curve. However care MUST be taken when selecting one that is appropriate, this specification defines a profile for using the BLS12-381 curve in (#ciphersuites) which as a curve currently achieves close to 128-bit security.
 
+## Security against the signer.
+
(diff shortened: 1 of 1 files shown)


tplooker (approved):
LGTM, thanks!

quartzjer (approved):

tplooker:
Multiple approvals, discussed on WG call, purely editorial non-normative content, merging

PR #87, opened 2022-03-21
[PROPOSAL] refactor key generation procedure

tplooker:
Initial proposal to refactor the key generation procedure to just rely upon the XOF rather than a combination of a HASH function and HKDF inline with @mikelodder7 proposal [here](https://github.com/decentralized-identity/bbs-signature/pull/84#pullrequestreview-911423521)

Outstanding questions
- Do we want to continue to support key_info in the key generation, this PR has removed it for now, however we can re-add it
- What guidance do we need to add to the cipher suite definition about calculating the appropriate value for this?


tplooker:
@BasileiosKal is currently reviewing the usage of HKDF as it would apply here including whether we want to continue to support `key_info` and if so, how.

andrewwhitehead:
This will likely be the same for any XOF -> Scalar generation in the spec like the challenge computation. I think we may want to add a generic function to the spec for this (HashToField for instance), implemented over the specific curve parameters.

The algorithm is probably more like this...

    salt = "BBS-SIG-KEYGEN-SALT-"
    h = XOF(salt || IKM)
    SK = 0
    while SK == 0:
        OKM = h.read(okm_key_gen_byte_length)
        SK = OS2IP(OKM) mod q
    return SK

tplooker:
Thanks @andrewwhitehead updated the key gen procedure proposal to be inline with the above.

mikelodder7 (approved):

tplooker:
Please hold off merging this, more detail on the outstanding things to discuss is outlined above.

PR #88, opened 2022-03-21
[PROPOSAL] generalize map message to scalar operation

tplooker:
Given we use this mapping procedure in the protocol to map values other than messages to scalars, this PR proposes to generalize that operation so we can use it elsewhere (such as for the challenge value as raised by @mikelodder7 [here](https://github.com/decentralized-identity/bbs-signature/pull/84#pullrequestreview-911423521), the alternative way we could do this instead is to use the underlying function of `hash_to_field` in places like this and leave the MapMessageToScalar as is.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -591,29 +591,32 @@ Procedure:
 3. return generators
 ```
 
-### MapMessageToScalar
+### MapValueToScalar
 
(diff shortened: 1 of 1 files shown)


tplooker:
Discussed on the WG, we are going to instead leave the definition for MapMessageToScalar as is and just use hash_to_field in required places directly. Therefore dis-regard this PR

tplooker:
Closing this PR now, have opened issue #89 to track the progress of resolving the original issue that brought this PR about in #84

Issue #78, opened 2022-03-07
Domain separation of generated signatures


tplooker:
At the moment within the scope of this specification there are several points of extension, such as

- How messages are mapped to their appropriate scalar values
- How the generators used to commit each message are created (e.g from a global, issuer specific or signature specific seed)

Many of these above decisions will be limited or at least formally defined in a concrete [ciphersuite](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-bls12-381-ciphersuite) to promote interoperability. However with multiple possible cipher suites overtime that make different choices, how do we ensure that a signature generated under one cipher suite is not confused with one generated under another cipher suite. 

One proposal to limit the possibility of this occurring is to define that the first message in a BBS signature MUST always be revealed further more this message could contain a range of information OR something more limited like a cipher suite identifier. As raised by @BasileiosKal on the call, putting too much information in this revealed message is likely to create a correlation risk that would erode some of the benefit that BBS signatures provide, so perhaps a better middle ground would be to instead encode the cipher suite identifier (e.g for the current BLS12381 cipher suite in the spec this could be the hash scalar produced from the string `BBS_BLS12381G1_XOF:SHAKE-256`)

mikelodder7:
If you do this approach you can no longer link messages across signatures anymore. The better approach which the original paper says is to use unique generators per signer.

tplooker:
@mikelodder7 can you please elaborate on how this is the case, the preferred proposal is merely to have the first message be a protocol identifier? I dont see how this would prevent cross signature?

tplooker:
Local generators would still not fully address the security concern described above, for example as a holder or verifier I would not know which message mapping methods are being used in the cipher suite so a malicious party could use this to attack in certain scenarios

mikelodder7:
If each message is mapped differently to a scalar per issuer then that means my name will not map to the same value in two different signatures and therefore I can't prove they are equal.

mikelodder7:
If you want a correlating identifier then why even use BBS+?!?

mikelodder7:
It solves the problem because local generators are unique per signer therefore the commitments are different and serve the same purpose as the cipher suite. This issue doesn't describe any attack @tplooker.

tplooker:
[...]

Issue #85, opened 2022-03-16
Correspondence between proof and public key


BasileiosKal:
In general, I was under the impression that a signature can be used to derive proofs that would be validated with only 1 specific key (the issuers PK).

That does not seem to be the case. The holder can present proofs for credentials signed from an issuer with public key `PK`, that will be validated with `PK * y = P2 * (x*y)`, where `y` a scalar chosen by the holder. My question is: Is this dangerous?? It is not really a forgery. It may compromise the issuer's and/or holder's accountability though, i.e., an issuer may falsely claim that they never signed a specific credential or a holder that they never possessed (and used to create an spk) a credential signed from that issuer.

You can find the detailed mathematics here: https://hackmd.io/@Vasileios/HyuXt_kf9
  and a reproduction here: https://github.com/BasileiosKal/dif-bbs-issue

In general the method is the following; Let the holder having the signature `(A, s, e)` on some messages. Let also the holder wanting to create a proof that would be validated with `PK * y` instead of `PK`, for some scalar `y`. During spkGen they will calculate `Abar_f = A’ * y * e + B * r1 * y` and `D_f = B * r1 * y + H0 * r2`. Note that `e(Abar_f, P2) = e(A', PK * y)`. Changing some additional values in the proof, the holder can present the spk; `(A’, Abar_f, D_f, …)` that will be validated with `PK * y`. 

A similar method works with a signature. The holder can present `(A*(y^-1), s, e*y)` which is a valid signature for the key `PK * y` (which is why the above method doesn't go against the security proofs given in the literature, i.e., a bbs proof, proves knowledge of a signature not only from the secret key `x` but also for a signature from the "secret key" `y * x`).

If a mitigation is needed, maybe including the issuer's PK as one of the signed messages will be an efficient solution, that solves the issuer’s and/or holder’s accountability issue.

mikelodder7:
The PK is known by the verifier so if the holder does this then it won’t validate properly

tplooker:
The fact this computation on the public key exists without a mitigation also degrades the non-repudiable nature of signatures also (e.g an issuer could use it as a way to say "i didn't sign the signature used to generate this spk")

tplooker:
Where a valid mitigation could be @mikelodder7's suggestion above.

BasileiosKal:
> The PK is known by the verifier so if the holder does this then it won’t validate properly

[...]

Issue #70, opened 2022-02-23
Simpler SpkGen option


andrewwhitehead:
I'm curious what's wrong with this simpler SPK alternative which doesn't specifically use `h_0` and `s`, treating them as a hidden message: https://hackmd.io/8t8o56mvTL6p-VZyayLnTA?view

One benefit here is that blind signatures become a true extension of the core spec, probably using a [named generator](#68), and signatures which don't involve this generator at all can be supported.

BasileiosKal:
That is truly very interesting. How will you prove zero knowledge though?? Following the same logic from the [paper](https://eprint.iacr.org/2016/663.pdf) doesn't seem possible, especially for `Abar` (the issuer will have to publish `Gbar ^ (-e)` for every signature, but the simulator will not know which one to use since they don't know the signature. Note that in the papers version the issuer publishes `Gbar ^ x` which is the same for all signatures)

andrewwhitehead:
At first glance, I think you just take a random `p` and `D`, set `A' = g^_1*p` and `Abar = g^_2*p - D`?

BasileiosKal:
Wright, that makes sense. In general, although I do agree that we may need to look more into it, this is a strong +1 from me. It simplifies spkGen so much that I think it is worth moving away from the papers. We will need to describe the simulator/ extractor in the spec though. IMO it's worth it.

Also, I think `C2` in the prover should be
```
C2 = D*r3~ - sum_over_R(H_i * m~_i)
```
And then in the verifier it should be
```
C2 = (-P1 - sum_over_D(H_i * m_i)) * c + D * r3^ - sum_over_R(H_i * m^_i)
```
(we may want to invert all the + and - for efficiency. However, i think the current equations for `C2` need some update like the above)

mikelodder7:
@andrewwhitehead prove that this is secure. If you're going to deviate from the original paper spec then please provide a proof of security.

andrewwhitehead:
It's actually trivial to convert from this proof output to the current proof output and still have it verify, which tells me it must be secure from the verifier's perspective. Is there more to the original proof than the Lemma 2 paragraph?

In any case, for a proof from this method `(A', Abar, D, e^, r2^, m^_0, {m^_i})` one can just set `Abar2 = Abar + D`, `r3^ = - r2^`, `s^ = m^_0` and output `(A', Abar2, D, e^, s^, r2^=0, r3^, {m^_i})`. A prover could do this now and save a few cycles.

[...]

PR #95, opened 2022-04-09
Add signature dst

BasileiosKal:
Closes Issues #78 and #85 

### Changes:
1. Adds the number of the revealed messages, the generators, the public key and some ciphersuite specific information to a signature domain separation tag (a signed -always revealed- message).
2. Switches notation from H0 to H_s, as discussed in the WG call

### Note:
This PR leaves the cipher suite specific info as optional for the `sig_dst`. The reason is that when it comes to the generators, the total number of messages etc., there are real risks for not adding them to the `sig_dst`.

The last one (total number of messages) is not documented in an issue to my knowledge, but I think it is well understood that it is dangerous to not include it. For example, the prover could reveal an additional “hidden” attribute by calculating `C1_new = C1 + H_j(U+1) * m~_j(U+1)` and `m^_j(U+1) = m~_j(U+1)` and have the proof be validated ([reproduction](https://github.com/BasileiosKal/dif-bbs-signature-dst-pr/blob/main/src/main.rs)). 

(Including the generators also solves this issue, so we could potentially only include the generators and not the total number of messages. That said, IMO it would be better to be more careful since it does not hurt performance or complexity)

In the case of the ciphersuite info though, IMO its good security practice to include it but not necessary?? I’m more than happy to update the pr and make the ciphersuite info mandatory, but to me it seems that the separation between ciphersuites could also happen elsewhere more naturally (in hash-2-curve for example, similarly to what the bls draft does)


draft-bbs-signatures.md
```diff
@@ -346,18 +346,27 @@ Procedure:
...
 Sign computes a signature from SK, PK, over a vector of messages. This method
 describes deterministic signing. For threshold signing, XOF can be replaced
-with a PRF due to the insecurity of deterministic threshold signing.
+with a PRF due to the insecurity of deterministic threshold signing. 
+
+The method takes as an input a list of the generators (H_s, H_d, H_1, ..., H_L). The first generator from that list (H_s), is used for the blinding value (s) of the signature. The second generator (H_d), is used to sign the signature's domain separation tag (sig_dst), which binds both signature and proof to a specific domain. Finally, the rest of the generators (H_1, ..., H_L), are used for the signed messages.
 
 ```
-signature = Sign(SK, PK, (msg_1,..., msg_L), (H_1,..., H_L))
+signature = Sign(SK, PK, (msg_1,..., msg_L), (H_s, H_d, H_1, ..., H_L))
```
tplooker, line 354: Im still pondering whether we make H_s and H_d cipher suite defined values rather than picking two generators and setting these as global fixed generators for all cipher suites, this would reduce this API's complexity by not having to supply them.

draft-bbs-signatures.md
```diff
@@ -346,18 +346,27 @@ Procedure:
...
-- H_1,..., H_L, points of G1. Generators used to sign the messages.
+- H_s, H_d, H_1,..., H_L, points of G1. Generators used to create the signature. H_s is
+                          used for the signature blinding value, H_d for the signature 
+                          domain separation tag and the rest for the signed messages.
 - SK, a secret key output from KeyGen
 - PK, a public key output from SkToPk
 
+Parameters:
+
+- CipherInfo, an optional string containing ciphersuite specific information.
```
tplooker, line 367: What makes cipherinfo a parameter and not an input here?

draft-bbs-signatures.md
```diff
@@ -366,37 +375,44 @@ Procedure:
 
 1. (W, H0, H) = octets_to_point(PK)
 
-2. h = XOF(SK  || msg[i] || ... || msg[L])
+2. sig_dst = HASH(PK || L || H_s || H_d || H_1 || ... || H_L || CipherInfo)
```
tplooker, line 378: Need to specify the form of `PK` and other parameters e.g octets_to_point

tplooker: Apologies, ignore saw the definition above r.e PK is the existing definition for L etc?

draft-bbs-signatures.md
BasileiosKal: Agreed. Updated the PR to define two more generator seeds for `H_s` and `H_d`. I also prefer this way. Since those generators are no longer calculated exactly the same as the message generators we may not need to pass them as input, as it was raised to the WG call.

draft-bbs-signatures.md
BasileiosKal: My thinking was that if cipherinfo is optional (which i think it should), that can be cached for different calls of the operations using the same ciphersuite, IMO it is suited for a parameter. This way, the inputs are only the values necessary for the core "mathematical" operations (in contrast, cipherinfo is only used for the sig_dst). This is similar to what `keyGen` does with `key_info`.

draft-bbs-signatures.md
BasileiosKal: That's a good point. L is implicitly defined as the total number of signed messages (based on that the input to sign is (msg_1, ..., msg_L)). I did notice though that the definition of L in terminology is different. Updated the definition of L to "the total number of signed messages" (there are still some inconsistencies with the definition of L even beyond this PR. I will submit a different PR to fix them).

draft-bbs-signatures.md
```diff
@@ -346,18 +351,23 @@ Procedure:
...
 
 Inputs:
 
-- msg_1,...,msg_L, octet strings. Messages to be signed.
+- msg_1,..., msg_L, octet strings. Messages to be signed.
 - H_1,..., H_L, points of G1. Generators used to sign the messages.
 - SK, a secret key output from KeyGen
 - PK, a public key output from SkToPk
 
+Parameters:
```
tplooker, line 366: Do h_s and h_d now have to listed as parameters here?

draft-bbs-signatures.md
tplooker: Ok understood and is consistent as you said with keyGen

draft-bbs-signatures.md
tplooker: Ok but do we need to convert L to bytes here? using I2OSP? Same for the other parameters too we need to make sure we fully define the form they are in to compute this hash

BasileiosKal: That's true. Note that the same consideration holds for the challenge hash in spkGen/verify. Also, the current approach is consistent with what the [Schnorr NIZK proofs RFC](https://datatracker.ietf.org/doc/html/rfc8235#section-2.3) does. That said, i would like to be more specific but i don't really like adding `point_to_octets` for every argument (for aesthetic reasons only). Another option could be a note that specifies the correct transformations for each element to be hashed (`point_to_octets` for points, `I2OSP` for numbers). I added such note in the definition to the HASH function in the Parameters section for review.

draft-bbs-signatures.md
BasileiosKal: IMO not technically since they are defined in the terminology and are not optional. We could add them for clarity in the parameters also though.

tplooker: Ok so we are saying that parameters differ from inputs in that they are optional? IMO I think it is clearer to have them defined here also.

draft-bbs-signatures.md
tplooker: I like your current approach which is to make the conversion implicit in the definition of the HASH function.

tplooker: I also noted this recommendation in the RFC you referenced is this something we should consider?

```
Within the hash function, there must be a clear boundary between any
   two concatenated items.  It is RECOMMENDED that one should always
   prepend each item with a 4-byte integer that represents the byte
   length of that item.  OtherInfo may contain multiple subitems.  In
   that case, the same rule shall apply to ensure a clear boundary
   between adjacent subitems.
```

BasileiosKal: Thank you! Yes I think we should. I opened up an [issue](https://github.com/decentralized-identity/bbs-signature/issues/96) to discuss some of the options. I will make another PR to move the spec to this direction.

draft-bbs-signatures.md
BasileiosKal: Done.

draft-bbs-signatures.md
BasileiosKal: See PR #97 for a proposal moving towards this direction. The addition to the definition of HASH is removed from this PR and re-introduced extended to #97.

mikelodder7 (commented):
I’ll need to work this in detail to understand why you can manipulate it to add an additional hidden message and still have it verify. I don’t think this is necessary until I convince myself. This seems superfluous. Being able to just arbitrarily add like this seems like the scheme itself is broken

BasileiosKal:
> I’ll need to work this in detail to understand why you can manipulate it to add an additional hidden message and still have it verify. I don’t think this is necessary until I convince myself. This seems superfluous. Being able to just arbitrarily add like this seems like the scheme itself is broken

I'm more than happy to discuss this further. Personally, i don't think that this breaks the scheme. The original proofs of security implicitly require the verifier to trust not only the generators but their number as well (i.e., the specific generators used for this specific signature), which is why IMO they are either in the signers PK or supplied by a trusted third-party CRS authority etc. Creating an arbitrary number of generators from a seed broke the trust to their number and this proposal re-enforces it.

Of-course this is just my intuition. Really keen to discuss.

draft-bbs-signatures.md
BasileiosKal: Thank you! I added some text to reflect that both H_s and H_d are mandatory, in their definitions in Terminology. Does this address your review?

andrewwhitehead: In regards to adding the length to hashed items, I do think it is necessary in a couple cases but not for all elements.

The current formula is `sig_dst = HASH(PK || L || H_s || H_d || H_1 || ... || H_L || CipherInfo)`. Here the encoded length of the PK and the generator points is known for the associated curve, and `L` serves to disambiguate the number of following generator points (perhaps it should be L+2 here). Prepending CipherInfo with its length helps to avoid any length extension attacks (although I don't believe SHAKE-256 suffers from that issue). If we added ApplicationInfo after CipherInfo, then adding a length to each would also help avoid any confusion around the boundaries between the two.

BasileiosKal: Thank you! Agreed. IMO it should be scalars (like the messages), integers (like L) and generic octet strings (like cipherInfo) that need to be prepended with their length (see PR #97 and issue #96 also). Regarding L specifically i do agree it may not be strictly necessary in `spk_dst` since it comes after PK and is followed by H_s, both of which have well defined lengths, but i think in general non-negative integers should be prepended with their length as a good practice.

(note that this is not only to avoid length extension which even if it was possible i don't think it is that dangerous against `spk_dst`, since the hash value would most likely change under this attack)

I also think i prefer `L` to `L+2` for aesthetic reason mainly and to more clearly reflect the number of signed messages. Since `H_s` and `H_d` are always supplied, essentially what might change is `L` either way.

andrewwhitehead:
> If sig_dst is 0 the sign will abort. We could use an XOF and keep drawing bytes instead of aborting but then verifier and holder should do the same procedure. I think that getting a 0 is so unlikely that just aborting would do fine while not raising the complexity of the draft. That said introducing a “hash_to_scalar” function (as I plan to), could also solve this issue without needing to abort.

Whether we abort or keep reading from the XOF, it would be nice to use the same method for both this calculation and the presentation challenge. I don't think that either case risks exposing sensitive information through a timing attack (maybe to some third party monitoring the process, but not to the verifier who will learn the hash inputs anyway?), so rather than aborting the program with a minuscule probability or adding special handing for the error case it seems like simply looping to the next hash output would be the best solution.

draft-bbs-signatures.md
andrewwhitehead: Scalars would also be encoded with a fixed length, and integers can be by defining the width of the field in the spec (f.ex. 4 bytes is plenty for the number of signed messages). For things like the revealed messages values in the challenge, that can be prepended by the number of following elements, each element being a pair of a 4-byte integer position and scalar value. In any case this is probably over-thinking things, but I suppose it doesn't hurt to be careful.

mikelodder7: We're getting into encoding layers and I don't think we should. That should be avoided. I2OSP and OSP2I are fine since they specify the length in bytes but that's all we should be using.

mikelodder7:
I don't think this is necessary. If you use the same info in the Fiat Shamir heuristic then this avoids the attack mentioned.

challenge = HASH(PK || H_s || H_1 || ... || H_L || C1 || C2 || T1 || T2 || nonce)

andrewwhitehead:
Adding the generators to the challenge hash is sufficient, but adding them to the DST is slightly more efficient for the prover, as they can cache the DST scalar and avoid hashing the generators with every presentation. The generators are still included in the Fiat-Shamir hash by virtue of the revealed DST message being hashed. If we agree that a DST message is necessary then I think that is the best place for them.

draft-bbs-signatures.md
tplooker: @mikelodder7 we have to sufficiently describe the encoding of elements as they pertain to operations within the scheme to promote interoperability

tplooker:
> but adding them to the DST is slightly more efficient for the prover, as they can cache the DST scalar and avoid hashing the generators with every presentation.

+1 that is my perspective also

mikelodder7:
I disagree. We're adding more work for both sides. How is that saving anything?!? I think we should arrange a time to discuss it. Github isn't the best place for this.

draft-bbs-signatures.md
BasileiosKal: I also feel that describing the formats of the messages to be hashed would be advantageous and is consistent with what other specs that deal with this stuff do. But I do also agree that going into too much detail and be too much restrictive is not a good thing. The minimum specification to achieve security must be preferred.

BasileiosKal: > Scalars would also be encoded with a fixed length, and integers can be by defining the width of the field in the spec (f.ex. 4 bytes is plenty for the number of signed messages). For things like the revealed messages values in the challenge, that can be prepended by the number of following elements, each element being a pair of a 4-byte integer position and scalar value. In any case this is probably over-thinking things, but I suppose it doesn't hurt to be careful.

When it comes to integers I agree and i updated PR #96 accordingly. When it comes to scalars there is two options i see,

1. Their length will be defined by the ciphersuite. This means that confusion regarding the used ciphersuite may lead to confusion regarding the scalars length.
2. The spec will define a fixed xLen to be used by all ciphersuites as an input to I2OSP.

IMO 2. seems too restrictive. Also spec extensions may define a different value for xLen which would lead to the same problem as option 1. For that reason, i prefer 1. and the proposed approach to prepend the lengths to scalars, although i don't feel that strongly about it and i would also go with option 2. or another proposal.

andrewwhitehead: If there is confusion over the ciphersuite being used then there would be big problems, that must be validated either through the DST or the generator creation process. I think we just say that scalars have a fixed length encoding defined by the ciphersuite.

tplooker:
> I disagree. We're adding more work for both sides. How is that saving anything?!? I think we should arrange a time to discuss it. Github isn't the best place for this.

Just adding my perspective here as I wont be able to attend the special call on this topic. The proposal is to include these public parameters (generators, total number of signed messages etc) in the signature DST instead of in the challenge computed for the proof. This is slightly more efficient as @andrewwhitehead said because the signature DST value could be cached across multiple proofs where the challenge value can not. Also these public parameters are set by the issuer so the safest way to prevent these from being manipulated, by the holder or any other party, is to just protected them in the original signature produced by the issuer via the signature DST.

draft-bbs-signatures.md
```diff
@@ -366,21 +378,25 @@ Procedure:
 
 1. (W, H0, H) = octets_to_point(PK)
 
-2. h = XOF(SK  || msg[i] || ... || msg[L])
+2. sig_dst = OS2IP(HASH(PK || L || H_s || H_d || H_1 || ... || H_L || CipherInfo)) mod q
```
tplooker, line 381: Can we also add in the ciphersuite ID here too? as the first argument? e.g the unique string that identifies the cipher suite

BasileiosKal: Done! Just a note, Ciphersuite_ID is not currently defined in ciphersuites. Will make an additional PR to add the definition.

draft-bbs-signatures.md
BasileiosKal: That's true. Well personally I would prefer even when ciphersuites define the length of the scalar, to be more careful and also enforce it here even if it is redundant, but im fine with the alternative. I will also be very interested in @christianpaquin input on that issue, since UProve seems to have taken this route.

draft-bbs-signatures.md
```diff
@@ -346,18 +351,26 @@ Procedure:
...
 
-- msg_1,...,msg_L, octet strings. Messages to be signed.
+- msg_1,..., msg_L, octet strings. Messages to be signed.
 - H_1,..., H_L, points of G1. Generators used to sign the messages.
 - SK, a secret key output from KeyGen
 - PK, a public key output from SkToPk
 
+Parameters:
+
+- CipherInfo, an optional string containing ciphersuite specific information.
```
tplooker, line 368:
```suggestion
- CipherInfo, an optional octet string containing ciphersuite specific information.
```

draft-bbs-signatures.md
christianpaquin: > I will also be very interested in @christianpaquin input on that issue, since UProve seems to have taken this route.

Just one thing I'd like to point out: currently the `H_d` generator is multiplied with a random `s` value. In U-Prove, we set that message value to the hash of various elements (issuer params, and an always-disclosed "token information" attribute). This allows issuers to encode a message that should always be disclosed (e.g., validity period, signature metadata, usage restrictions, etc.) when presenting the signature, without "wasting" another message for that purpose if needed. Maybe something to consider here?

draft-bbs-signatures.md
```diff
@@ -366,21 +379,25 @@ Procedure:
...
-2. h = XOF(SK  || msg[i] || ... || msg[L])
+2. sig_dst = OS2IP(HASH(PK || L || H_s || H_d || H_1 || ... || H_L || Ciphersuite_ID || CipherInfo)) mod q
+
+3. if sig_dst is 0, abort
 
-3. for rand_el in (e, s) do
+4. h = XOF(SK  || msg[i] || ... || msg[L])
 
-4.      rand_el = OS2IP(h.read(64)) mod q.
+5. for rand_el in (e, s) do
```
tplooker, line 388: side note that tripped me up here rand_el implies that the resulting values for e and s are entirely random when they are not should we name this something else?

draft-bbs-signatures.md
```diff
@@ -411,13 +436,15 @@ Procedure:
 
 4. if KeyValidate(pub_key) is INVALID
 
-5. B = P1 + H0 * s + H_1 * msg_1 + ... + H_L * msg_L
+5. sig_dst = OS2IP(HASH(PK || L || H_s || H_d || H_1 || ... || H_L || Ciphersuite_ID || CipherInfo)) mod q
```
tplooker, line 439: Considering we are talking about having the parmeters suitable implicitly converted would the same make sense to the ouput of the HASH function e.g always returns the form of OS2IP?

draft-bbs-signatures.md
BasileiosKal: Thank you very much for your input! That is interesting. In bbs+ the `s` value is never revealed (the paper goes to great lengths actually to keep it secret, see issue #70  for some context). This is to avoid correlation. And from the conversation in the issue i linked to, it seems that there is no getting around the fact that we need `s` to be random and we need it to be secret. That said, it is a very interesting idea and definitely worth thinking more about it.

draft-bbs-signatures.md
BasileiosKal: Thank you! That's a good point. I opened PR #101 which is towards this direction and if adopted will replace the OS2IP(HASH(...)) here.

draft-bbs-signatures.md
```diff
@@ -112,8 +112,11 @@ generator
 H\[i\]
 : The generator corresponding to a given msg.
 
-H0
-: A generator for the blinding value in the signature.
+H_s
+: A generator for the blinding value in the signature. H_s is defined by each ciphersuite must always be supplied to the operations listing it as a parameter.
```
tplooker, line 116:
```suggestion
: A generator for the blinding value in the signature. The value of H_s is defined by each ciphersuite and must always be supplied to the operations listing it as a parameter.
```

draft-bbs-signatures.md
```diff
@@ -112,8 +112,11 @@ generator
...
+
+H_d
+: A generator for the signature domain separation tag (sig_dst), which binds both signature and proof to a specific domain. H_d is defined by each ciphersuite and must always be supplied to the operations listing it as a parameter.
```
tplooker, line 119:
```suggestion
: A generator for the signature domain separation tag (sig_dst), which binds both signature and proof to a specific domain. The value of H_d is defined by each ciphersuite and must always be supplied to the operations listing it as a parameter.
```

draft-bbs-signatures.md
tplooker: How about `element` or `item` instead?

draft-bbs-signatures.md
```diff
@@ -397,6 +414,14 @@ Inputs:
 - signature, octet string.
 - PK, a public key in the format output by SkToPk.
 
+Parameters:
+
+- CipherInfo, an optional string containing ciphersuite specific information.
```
tplooker, line 419:
```suggestion
- CipherInfo, an optional octet string containing ciphersuite specific information.
```

draft-bbs-signatures.md
```diff
@@ -438,6 +465,14 @@ Inputs:
 - signature, octet string in output form from Sign
 - pm, octet string
 
+Parameters:
+
+- CipherInfo, an optional string containing ciphersuite specific information.
```
tplooker, line 470:
```suggestion
- CipherInfo, an optional octet string containing ciphersuite specific information.
```

draft-bbs-signatures.md
```diff
@@ -544,6 +581,14 @@ Inputs:
 - RevealedIndexes, vector of unsigned integers. Indexes of revealed messages.
 - pm, octet string
 
+Parameters:
+
+- CipherInfo, an optional string containing ciphersuite specific information.
```
tplooker, line 586:
```suggestion
- CipherInfo, an optional octet string containing ciphersuite specific information.
```

BasileiosKal:
Some more updates,
-	All suggestions made by @tplooker have been integrated.
-	Also proposed by @tplooker, `sig_dst` is passed to the XOF used to calculate `e` and `s`. This has the advantage that signatures on the same messages but under different cipher suites (using the same generators) or application domains, will also be different, which is nice.
-	`CipherInfo` is changed to `ApplicationInfo`. I like this better since it captures separation between blind and non-blind signatures. Chiphersuite separation will most probably happen with `Ciphersuite_ID` either way. Currently I don’t see any requirement to pass `CipherInfo` in the `sig_dst` but if anyone has a relative use case where both `Ciphersuite_ID` and `CipherInfo` will be useful I will be happy to update the PR.
-	If `ApplicationInfo` gets adopted, some guidance will be needed on what this value should be.

andrewwhitehead:
I was imagining different use cases for CipherInfo and ApplicationInfo. CipherInfo would represent standard revealed parameters as required by the cipher suite, while ApplicationInfo would be more of a private extension method for use cases that build on an existing cipher suite.

With [ldp-bbs2020](https://w3c-ccg.github.io/ldp-bbs2020/#required-reveal-statements) as an example, the CipherInfo would include at least the `verificationMethod` and `proofPurpose` attributes, no longer encoded as separate messages. Each cipher suite would need to define the procedure for generating an octet string from its parameters, if any, for example it might use `bytes(canonical_json({"verificationMethod": ..., "proofPurpose": ...}))` in this case. Revocation registry details might be another example of cipher suite-defined parameters that would need to be encoded in this way so that the holder cannot choose not to reveal them.

ApplicationInfo might correspond to the token information field in U-Prove, allowing a way to formalize additional contextual parameters about how the signature can be applied and interpreted:

> The token information field is used to encode token-specific information that is always disclosed to Verifiers, such as token usage restrictions, a validity period (see Section 3.3), or token metadata.

As an example, an issuer might use ApplicationInfo to restrict the presentation of a signature to a specific IP address to which it was issued (this wouldn't be very privacy-preserving). However, since there is no application identifier to restrict the interpretation of ApplicationInfo, it might be better to forget about that concept and require the cipher suite to define these parameters if they are to be supported. If we are only encoding one such parameter then ApplicationInfo seems like a better name.

BasileiosKal:
That's very interesting. I'm kinda conflicted because from the one hand it would be nice for ciphersuites to have the ability to define "always revealed messages" but on the other hand i feel it breaks a little the separation between ciphersuite and application layer. For example, a representation layer specification like ldp-pps2020 would have to define a (cryptographic layer) ciphersuite and i don't think its wise (and fair) to ask from the developers of those specifications to have to define a cryptographic suite.

I wonder if for a spec like ldp-bbs2020 would be enough to use `ApplicationInfo` to force certain messages to be revealed. That said, since it seems that there may exist cases that something like that would be useful, i think it would be better to add it.

Also regarding `ApplicationInfo`, i think it would not be hard to give some requirements regarding its contains and meaning, at least to avoid privacy issues. Perhaps `DomainInfo` would be a better name?

tplooker:
This raises a good point for me and I tend to resonate with this POV, essentially could we have both one is a parameter "CipherInfo" set by the ciphersuite, the other is a parameter that is supplied to the sign, verify, spkGen and spkVerify operations? I can see two different possible applications for ApplicationInfo

1. Usecases where the signer wants to include information that must always be revealed in the resulting proof
2. Usecases where no selective disclosure is even required, instead all the signed information must always be revealed and instead the scheme is really just been used more for just its group signature properties.

If two resonates with folks then I wonder if the messages and generator parameters become optional in this setting?

tplooker:
The only thing that doesn't quite fit for me is whether ApplicationInfo should form part of a domain separation tag? All other info that goes into the computation of sig_dst seems to fit this model nicely however ApplicationInfo does not? Perhaps thats just a mental model thing however as the only alternative I see is to commit the application info into the signature using a seperate generator which would be in-efficient?

tplooker:
Just to muse on how a proposal to support both the cipher info and application info might work I would suggest it would look  like the following

```
signature = Sign(SK, PK, (msg_1,..., msg_L), (H_1,..., H_L), ApplicationInfo)

Inputs:

- msg_1,..., msg_L, octet strings. Messages to be signed.
- H_1,..., H_L, points of G1. Generators used to sign the messages.
- SK, a secret key output from KeyGen
- PK, a public key output from SkToPk

Parameters:

- Ciphersuite_ID, octet string. The unique ID of the ciphersuite.
- CipherInfo, an optional octet string containing context and ciphersuite specific
                   information. If not supplied, it defaults to the empty octet string.
- H_s, point of G1. The generator for the blinding value of the signature.
- H_d, point of G1. The generator used to sign the signature domain separation tag.

Outputs:

- signature, an octet string

Procedure:

1. (W, H0, H) = octets_to_point(PK)

2. dst = (PK || L || H_s || H_d || H_1 || ... || H_L || Ciphersuite_ID || CipherInfo)

3. sig_info = OS2IP(HASH(dst || ApplicationInfo)) mod q

4. if sig_info is 0, abort

5. h = XOF(SK  || dst || msg[i] || ... || msg[L])

6. for element in (e, s) do

7.      element = OS2IP(h.read(64)) mod q

8.      if element = 0, go back to step 4

9. B = P1 + H_s * s + H_d * sig_dst + H_1 * msg_1 + ... + H_L * msg_L

10. A = B * (1 / (SK + e))

11. signature = (point_to_octets_min(A), e, s)

12. return signature
```

BasileiosKal:
I would argue that `CipherInfo` is not needed. This keeps a clear definition of what a ciphersuite is (the definition of a curve, a hash etc) which we should try to keep IMO. I wonder if we should have an additional section of `Optional Inputs` under the not optional ones. Something like,

```
signature = Sign(SK, PK, (msg_1,..., msg_L), (H_1,..., H_L))

Inputs:

- msg_1,..., msg_L, octet strings. Messages to be signed.
- H_1,..., H_L, points of G1. Generators used to sign the messages.
- SK, a secret key output from KeyGen
- PK, a public key output from SkToPk

Optional Inputs:

- ApplicationInfo, an optional octet string containing application specific
                    information. If not supplied, it defaults to the empty octet string.

Parameters:

- Ciphersuite_ID, octet string. The unique ID of the ciphersuite.
- H_s, point of G1. The generator for the blinding value of the signature.
- H_d, point of G1. The generator used to sign the signature domain separation tag.

Outputs:

- signature, an octet string

Procedure:

1. (W, H0, H) = octets_to_point(PK)
2. ...
```
This may seem as a lot of text before we define the procedure but,
1. This leaves the definition of each function as simple and as close to known APIs (i.e., ecdsa, rsa etc) as possible.
2.  People that don't care about `ApplicationInfo` can more easily ignore it.
3. Since the `Optional Inputs` and the `Parameters` are the same for all core functions (sign, verify, spkGen and spkVerify) we can move those operations in a separate section (like core signature operations??) and define them only once in the start of this new section.

I also think that naming the `ApplicationInfo`, `ProtectedInfo` instead, facilitates their usage for always revealed messages and for separation between blind and non-blind signatures, that in my opinion are the 2 main use cases for the additional information.

tplooker:
> I would argue that CipherInfo is not needed. This keeps a clear definition of what a ciphersuite is (the definition of a curve, a hash etc) which we should try to keep IMO. I wonder if we should have an additional section of Optional Inputs under the not optional ones

I still think CipherInfo is worth considering separately, this is distinct from the application level structure "header" IMO, CipherInfo is primarily defined to allow those who define new ciphersuites to have a somewhere to include additional public parameters they may define.

BasileiosKal:
An update to the PR based on the conversations of the WG around notation.
1.	The term `domain` is used instead of `sig_dst`.
2.	The term `header` is used for the extra info passed to the hash of the domain (ex `sig_dst`)
3.	The term `messages` is preserved for now instead of `payloads` since there wasn’t a clear consensus around moving to the `payloads` terminology yet?
4.	I took a first shot on describing the functionality of the signature’s domain in the Messages and generators subsection (in considerations) but this may need its own section

Note also that I don’t thing we necessarily need to wait for PRs #97 and #101 to be closed for this PR to be updated. Although its true that those PRs are related, if they are merged, a wider update in the spec may be needed, which would include the changes here, if that PR is merged.

Also, the `header` is passed as an input to each operation. To me this is not ideal but making it a parameter IMO doesn’t make much sense either, considering the always revealed messages use-case. I do plan to make another PR with an alternative proposal (like proposal [here](https://github.com/decentralized-identity/bbs-signature/pull/95#issuecomment-1101499779) for an `Optional Input` section) so i think it's fine to move with this approach for now.

tplooker:
@BasileiosKal can you resolve merge conflicts with this PR?

draft-bbs-signatures.md
```diff
@@ -366,36 +379,50 @@ Procedure:
 
 1. (W, H0, H) = octets_to_point(PK)
 
-2. h = XOF(SK  || msg[i] || ... || msg[L])
+2. Generators =  (H_s || H_d || H_1 || ... || H_L)
```
tplooker, line 382:
```suggestion
2. generators =  (H_s || H_d || H_1 || ... || H_L)
```

for case consistency

draft-bbs-signatures.md
```diff
@@ -454,43 +493,47 @@ Procedure:
 
 5. if KeyValidate(PK) is INVALID abort
 
-6. for rand_el in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU): 
+6. Generators =  (H_s || H_d || H_1 || ... || H_L)
```
tplooker, line 496:
```suggestion
6. generators =  (H_s || H_d || H_1 || ... || H_L)
```

for case consistency

draft-bbs-signatures.md
```diff
@@ -454,43 +493,47 @@ Procedure:
...
+
+7. domain = OS2IP(HASH(PK || L || Generators || Ciphersuite_ID || header)) mod q
+
+8. for rand_el in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU):
```
tplooker, line 500:
```suggestion
8. for element in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU): 
```

draft-bbs-signatures.md
```diff
@@ -454,43 +493,47 @@ Procedure:
...
 
-6. for rand_el in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU): 
+6. Generators =  (H_s || H_d || H_1 || ... || H_L)
+
+7. domain = OS2IP(HASH(PK || L || Generators || Ciphersuite_ID || header)) mod q
+
+8. for rand_el in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU): 
 
-7.      rand_el = HASH(PRF(8*ceil(log2(q)))) mod q
+9.      rand_el = HASH(PRF(8*ceil(log2(q)))) mod q
```
tplooker, line 502:
```suggestion
9.      element = HASH(PRF(8*ceil(log2(q)))) mod q
```

draft-bbs-signatures.md
```diff
@@ -454,43 +493,47 @@ Procedure:
...
 
-8.      if rand_el = 0, go back to step 7
+10.      if rand_el = 0, go back to step 7
```
tplooker, line 504:
```suggestion
10.      if element = 0, go back to step 7
```

draft-bbs-signatures.md
```diff
@@ -558,21 +609,25 @@ Procedure:
 
 4. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1,...,m^_jU)) = spk
 
-5. C1 = (Abar - D) * c + A' * e^ + H0 * r2^
+5. Generators =  (H_s || H_d || H_1 || ... || H_L)
```
tplooker, line 612:
```suggestion
5. generators =  (H_s || H_d || H_1 || ... || H_L)
```

tplooker (changes requested):
A couple more minor tweaks, getting there!

draft-bbs-signatures.md
```diff
@@ -404,36 +417,50 @@ Procedure:
...
 
-2. h = XOF(SK  || msg[i] || ... || msg[L])
+2. generators =  (H_s || H_d || H_1 || ... || H_L)
+
+3. domain = OS2IP(HASH(PK || L || generators || Ciphersuite_ID || header)) mod q
+
+4. if domain is 0, abort
 
-3. for rand_el in (e, s) do
+5. h = XOF(SK  || domain || msg[i] || ... || msg[L])
```
tplooker, line 426: I like doing element access via the square bracket syntax as you have done here, but is this not in consistent with line 434? I'd prefer we do it this way but it needs to be the same throughout the doc

draft-bbs-signatures.md
```diff
@@ -404,36 +417,50 @@ Procedure:
...
+result = Verify(PK, (msg_1,..., msg_L), (H_1,..., H_L), signature, header)
 
 Inputs:
 
 - msg_1,..., msg_L, octet strings. Messages in input to Sign.
 - H_1,..., H_L, points of G1. The generators in input to Sign.
 - signature, octet string.
 - PK, a public key in the format output by SkToPk.
+- header, an optional octet string containing context and application specific
+          information. If not supplied, it defaults to the empty string.
```
tplooker, line 457:
```suggestion
          information. If not supplied, it defaults to an empty string.
```

BasileiosKal: Updated. Thanks!

draft-bbs-signatures.md
tplooker:
```suggestion
5. h = XOF(SK  || domain || msg_1 || ... || msg_L)
```

draft-bbs-signatures.md
```diff
@@ -123,8 +123,11 @@ generator
 H\[i\]
 : The generator corresponding to a given msg.
 
-H0
-: A generator for the blinding value in the signature.
+H_s
```
tplooker, line 126:
```suggestion
Hs
```

draft-bbs-signatures.md
```diff
@@ -123,8 +123,11 @@ generator
 H\[i\]
```
tplooker, line 123: Can we remove this entry?

tplooker (approved):

andrewwhitehead (approved):

tplooker:
@BasileiosKal can you please resolve conflicts?

tplooker:
Open three weeks, several reviews and discussion on WG calls, multiple approvals now, merging

Issue #96, opened 2022-04-11
Element encoding when passed to HASH or XOF.


BasileiosKal:
See discussion in PR #95

In the spec, we need a way to pass elements of different types (EC points, integers, scalars etc) to a hash function. In that case, the encoding of those elements becomes important. IMO there are two options:
1. Explicitly making the transformation to octets for each hashed element. That would look something like
```
HASH(point_to_octets(PK), I2OSP(L), point_to_octets(H_1),..., point_to_octets(H_L),...)
```
2. Implicitly make the conversion (in the definition of the HASH for example or in a separate section etc.). This would look something like,
```
HASH(PK, L, H_1,..., H_L,... )
``` 
Personally I prefer 2. This would mean that we can also specify rules like the one Tobias pointed out [here ](https://github.com/decentralized-identity/bbs-signature/pull/95#discussion_r847082901)i.e., that each element should also be appended with its length, which i agree with. A consideration here is if we need to specify this rule for every element (even the point_to_octets outputs) or it is sufficient to specify it for general octet strings and non-negative integers (lengths of lists)??. IMO the latter will suffice, since there is not really any ubiquities that come to mind regarding the output of point_to_octets (which is also an octet string but with well defined length), but i could be mistaken.

BasileiosKal:
2 additional notes:

1. To @tplooker  [point](https://github.com/decentralized-identity/bbs-signature/pull/97#discussion_r847895296), making the points in G1 or G2 be prepended with their length optional swifts the weight to the cipher suite to describe when this will hold for interoperability. As a result IMO we should either make this a requirement or remove it entirely (i vote for the later, since the length of the `points_to_octets` output can be defined by its value).

2. An additional consideration is whether we need to also prepend the lengths of lists to be hashed, to the list them serfs (for example the list of generators, or in the list of disclosed messages etc.)

tplooker:
@BasileiosKal can you confirm with the merging of #97 whether we can now close this issue?

BasileiosKal:
Yeap! Closing

PR #97, opened 2022-04-11
Encodings for elements to be hashed

BasileiosKal:
First approach to close issue #96

Encodings suggestion of elements to be hashed (similar to the ones used by [U-Prove](https://github.com/microsoft/uprove-csharp-sdk/blob/master/docs/U-Prove%20Cryptographic%20Specification%20V1.1%20Revision%203.pdf)). 

This PR follows option 2 from the above issue, i.e., the encoding of elements to be hashed is described in a separate section and used implicitly throughout the document. See PR #95  for more context.

I also suggest that scalars, non-negative integers, and generic octet strings must always be prepended with their length while it is only recommended for EC points.


draft-bbs-signatures.md
```diff
@@ -231,6 +231,17 @@ In definition of this signature scheme there are two possible variations based u
 
 Throughout the operations of this signature scheme, each message that is signed is paired with a specific generator (point in G1). Specifically, if a generator `H_1` is raised to the power of `msg_1` during signing, then `H_1` should be raised to the power of `msg_1` in all other operations as well (signature verification, proof generation and proof verification). For simplicity, each function will take as input the list of generators to be used with the messages. Those generators can be any distinct element from the generators list `H`. Applications for efficiency can elect to pass the indexes of those generators to the list `H` instead. Care must be taken for the correct generator to be raised to the correct message in that case.
 
+### Encoding of elements to be hashed.
+To avoid ubiquities, each element passed to the HASH or the XOF function (either by itself or concatenated with other elements), must first be encoded to an appropriate format, depending on its type. Specifically,
```
tplooker, line 235:
```suggestion
To avoid ambiguity, each element passed to the HASH or the XOF function (either by itself or concatenated with other elements), must first be encoded to an appropriate format, depending on its type. Specifically, 
```

I believe ambiguity might be a better word here?

draft-bbs-signatures.md
```diff
@@ -231,6 +231,17 @@ In definition of this signature scheme there are two possible variations based u
...
+- Points in G1 or G2 must be encoded using `point_to_octets`.
+- Non-negative integers must be encoded using `I2OSP`. 
+- Scalars must be zero-extended to a length that is a multiple of 8 bits. Then, the extended value binary representation is encoded as an octet string.
+- Octet strings must be zero-extended to a length that is a multiple of 8 bits. Then, the extended value is encoded directly.
+
+After encoding, scalars, non-negative integers, and octet strings MUST be prepended with the length (number of bytes) of their binary representation (after they are encoded). The same is RECOMENDED for points in G1 or G2 as well. To encode the length of the above elements, the length's binary value must be zero-extended to 32 bits. The extended value binary representation is then encoded as an octet string. For example, the string `0x14d` will be encoded as `0x00000002014d`. Similarly, the number `236195248387` will be encoded as `0x0000000536fe557903`. Values larger than 2^32 - 1, MUST be rejected.
```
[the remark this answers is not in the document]
BasileiosKal: Thank you! I haven't considered that. IMO it would be part of the suites but it's not worth raising the complexity since the length of the `points_to_octets` output is determined but its value either way. The recommendation was just to be extra careful. Removed for know.

draft-bbs-signatures.md
BasileiosKal: Thanks! Updated.

draft-bbs-signatures.md
```diff
@@ -231,6 +231,17 @@ In definition of this signature scheme there are two possible variations based u
 
 Throughout the operations of this signature scheme, each message that is signed is paired with a specific generator (point in G1). Specifically, if a generator `H_1` is raised to the power of `msg_1` during signing, then `H_1` should be raised to the power of `msg_1` in all other operations as well (signature verification, proof generation and proof verification). For simplicity, each function will take as input the list of generators to be used with the messages. Those generators can be any distinct element from the generators list `H`. Applications for efficiency can elect to pass the indexes of those generators to the list `H` instead. Care must be taken for the correct generator to be raised to the correct message in that case.
 
+### Encoding of elements to be hashed.
+To avoid ambiguity, each element passed to the HASH or the XOF function (either by itself or concatenated with other elements), must first be encoded to an appropriate format, depending on its type. Specifically,
```
tplooker, line 235:
```suggestion
To avoid ambiguity, each element passed to the HASH or the XOF function, either by itself or concatenated with other elements, must first be encoded to an appropriate format, depending on its type. Specifically, 
```

draft-bbs-signatures.md
```diff
@@ -231,6 +231,17 @@ In definition of this signature scheme there are two possible variations based u
 
 Throughout the operations of this signature scheme, each message that is signed is paired with a specific generator (point in G1). Specifically, if a generator `H_1` is raised to the power of `msg_1` during signing, then `H_1` should be raised to the power of `msg_1` in all other operations as well (signature verification, proof generation and proof verification). For simplicity, each function will take as input the list of generators to be used with the messages. Those generators can be any distinct element from the generators list `H`. Applications for efficiency can elect to pass the indexes of those generators to the list `H` instead. Care must be taken for the correct generator to be raised to the correct message in that case.
 
+### Encoding of elements to be hashed.
+To avoid ambiguity, each element passed to the HASH or the XOF function (either by itself or concatenated with other elements), must first be encoded to an appropriate format, depending on its type. Specifically, 
+- Points in G1 or G2 must be encoded using `point_to_octets`.
```
tplooker, line 236:
```suggestion
- Points in G1 or G2 must be encoded using the `point_to_octets` implementation for a particular ciphersuite.
```

draft-bbs-signatures.md
```diff
@@ -231,6 +231,17 @@ In definition of this signature scheme there are two possible variations based u
...
+- Non-negative integers must be encoded using `I2OSP`. 
+- Scalars must be zero-extended to a length that is a multiple of 8 bits. Then, the extended value binary representation is encoded as an octet string.
+- Octet strings must be zero-extended to a length that is a multiple of 8 bits. Then, the extended value is encoded directly.
+
+After encoding, scalars and octet strings MUST be prepended with the length (number of bytes) of their binary representation (after they are encoded). To encode the length of the above elements, the length's binary value must be zero-extended to 32 bits. The extended value binary representation is then encoded as an octet string. For example, the string `0x14d` will be encoded as `0x00000002014d`. Similarly, the number `236195248387` will be encoded as `0x0000000536fe557903`. Values larger than 2^32 - 1, MUST be rejected.
```
tplooker, line 241:
```suggestion
After encoding, scalars and octet strings MUST be prepended with a value representing the length of their binary representation in the form of the number of bytes. This length's binary value must be zero-extended to 32 bits. The combined value (encoded value + length prefix) binary representation is then encoded as a single octet string. For example, the string `0x14d` will be encoded as `0x00000002014d`. Similarly, the number `236195248387` will be encoded as `0x0000000536fe557903`. Values larger than 2^32 - 1, MUST be rejected.
```

draft-bbs-signatures.md
```diff
@@ -231,6 +231,17 @@ In definition of this signature scheme there are two possible variations based u
...
+
+For readability, this document makes these transformations implicitly, but they MUST precede every call to the HASH or XOF function.
```
tplooker, line 243: Wonder to make this 100% clear whether we should give an example of this transform?

draft-bbs-signatures.md
BasileiosKal: Done.

draft-bbs-signatures.md
BasileiosKal: Done.

draft-bbs-signatures.md
BasileiosKal: Thanks! Updated this section following your suggestions.

draft-bbs-signatures.md
BasileiosKal: Added an example dealing with a "octet string" (an ascii message) and a number. There are other options to include (points etc.) but then we may want to move the example to an appendix since it will likely be more extensive.

BasileiosKal:
Updates proposed by @mikelodder7 and @andrewwhitehead  during the WG call of 18/ 4/2022:
1.	Lengths are extended to 8 bytes (for non-DST’s octet strings). Non-negative integers are also updated to be represented with 8 bytes, to be consistent with the representation of the lengths of octet strings (which are also non-negative integers). 
2.	The second paragraph is updated to make use of I2OSP (instead of zero-extensions).
3.	The example is updated accordingly.

mikelodder7 (approved):
LGTM now

tplooker (approved):

draft-bbs-signatures.md
```diff
@@ -231,6 +231,31 @@ In definition of this signature scheme there are two possible variations based u
 
 Throughout the operations of this signature scheme, each message that is signed is paired with a specific generator (point in G1). Specifically, if a generator `H_1` is raised to the power of `msg_1` during signing, then `H_1` should be raised to the power of `msg_1` in all other operations as well (signature verification, proof generation and proof verification). For simplicity, each function will take as input the list of generators to be used with the messages. Those generators can be any distinct element from the generators list `H`. Applications for efficiency can elect to pass the indexes of those generators to the list `H` instead. Care must be taken for the correct generator to be raised to the correct message in that case.
 
+### Encoding of elements to be hashed.
+To avoid ambiguity, each element passed to the HASH or the XOF function, either by itself or concatenated with other elements, must first be encoded to an appropriate format, depending on its type. Specifically,
```
tplooker, line 235: I think we need to clarify that the encoding rules defined below are per-element 

```suggestion
To avoid ambiguity, each element passed to the HASH or the XOF function, including situations when multiple elements are supplied in a concatenated form, must first be encoded to an appropriate format, depending on its type. Specifically,
```

BasileiosKal: Thanks! Updated.

andrewwhitehead (approved):

tplooker:
Multiple reviews and now approvals, open 10 days, discussed on latest WG call, merging

Issue #100, opened 2022-04-12
Add a security consideration for when the scalar representation of a message is zero


tplooker:
If a mapped message value outputted by the MapMessageToScalarAsHash evaluates to zero then effectively this message wont be signed, therefore we should add a security consideration to the spec to handle this case.

tplooker:
Addressed in #101, closing

PR #101, opened 2022-04-13
Adding a hash-to-scalar function

BasileiosKal:
This PR introduces a `hash_to_scalar` function as discussed in the WG call. The goal is for the function to replace some of the hashing to scalars happening internally in the operations, i.e., for the `e`, `s` calculation in `sign`, in the random elements in `spkGen` and potentially for the challenge value and the `sig_dst` if we decide to use that (see PR #95). The function is not inserted yet in those operations to not block #95. If that is resolved first i will update this PR.

### Note:
`hash_to_field` is removed with this PR (from `MapMessageToScalarAsHash`) and is replaced with `hash_to_scalar`. Note that this will also close issue #100 since `hash_to_scalar` checks if an element is 0. My reasoning is that `hash-to-scalar` serves the same functionality while been able to be used for the cases detailed above, without needing to worry about dst's etc. Note that aside from the use of dst, `hash-to-scalar` is almost exactly `hash-to-field` using `expand-message-xof` and with `m` (the fields extension degree) = 1. 

Also the dst is not removed from `MapMessageToScalarAsHash` (where it is useful). A note is also added stating that people can use `hash-to-field` if they want. However, i feel that the current construction is more flexible to serve the needs of the spec, but really keen to discuss alternatives and other proposals.


andrewwhitehead:
I'd prefer to see the lengths prepended in MapMessageToScalarAsHash, maybe like this?

    1. If len(dst) > 2^8-1 or len(msg) > 2^32-1, abort
    2. dst_prime = I2OSP(len(dst), 1) || dst
    3. msg_prime = I2OSP(len(msg), 4) || msg
    4. result = hash_to_scalar(dst_prime || msg_prime, 1)
    5. return result

draft-bbs-signatures.md
```diff
@@ -633,9 +633,50 @@ Outputs:
...
+
+3. msg_prime = msg || I2OSP(64, 2) || dst_prime
+
+4. result = hash_to_scalar(msg_prime, 1)
+
+5. return result
+```
+
+### Hash to scalar
+The following procedure describes how to hash an arbitrary octet string to `n` scalar values in the multiplicative group of integers mod q. This procedure acts as a helper function, and it used internally in various places within the operation described in the spec. To map a message to a scalar that would be passed as input to the [Sign](#sign), [Verify](#verify), [spkGen](#spkgen) and [spkVerify](#spkverify) functions, one must use [MapMessageToScalarAsHash](#mapmessagetoscalar) instead.
```
tplooker, line 648:
```suggestion
This operation describes how to hash an arbitrary octet string to `n` scalar values in the multiplicative group of integers mod q. This procedure acts as a helper function, and it used internally in various places within the operation described in the spec. To map a message to a scalar that would be passed as input to the [Sign](#sign), [Verify](#verify), [spkGen](#spkgen) and [spkVerify](#spkverify) functions, one must use [MapMessageToScalarAsHash](#mapmessagetoscalar) instead.
```

BasileiosKal: Thanks! Done.

BasileiosKal:
> I'd prefer to see the lengths prepended in MapMessageToScalarAsHash, maybe like this?

Thank you! I updated the PR based on your suggestion.

IMO this brings also an interesting point for PR #97, i.e., it makes little sense to have a dst be appended with a 4 bytes length (as #97 suggests). Maybe an "unless otherwise stated" is needed.

andrewwhitehead:
Yes adding the DST length seems to just add an unnecessary restriction in this case.

BasileiosKal:
That's true. Just to make sure i understand. Is your preference to remove the addition of the DST length from this PR or remove/update the descriptions of PR #97.

If it is the first case, maybe a better option would be to remove the lengths of both the message and the dst from `MapMessageToScalarAsHash` and just point to the recommendation in PR #97.

BasileiosKal:
Update: I added the dst handling in PR #97, so now the two PRs are consistent. If we decide to go that route, the only thing that will neet to be updated IMO is noting that the transformation descried in #97 ..."-**may**- happen implicitly, but they MUST precede every call to the HASH or XOF function."

BasileiosKal:
Following the [updates](https://github.com/decentralized-identity/bbs-signature/pull/97#issuecomment-1104519302) on PR #97, extended the length of a msg to be mapped to a scalar as hash to 8 bytes max.

draft-bbs-signatures.md
```diff
@@ -633,9 +633,50 @@ Outputs:
...
+```
+
+### Hash to scalar
+This operation describes how to hash an arbitrary octet string to `n` scalar values in the multiplicative group of integers mod q. This procedure acts as a helper function, and it is used internally in various places within the operations described in the spec. To map a message to a scalar that would be passed as input to the [Sign](#sign), [Verify](#verify), [spkGen](#spkgen) and [spkVerify](#spkverify) functions, one must use [MapMessageToScalarAsHash](#mapmessagetoscalar) instead.
+
+The `hash_to_scalar` procedure hashes elements using an extendable-output function (XOF). Applications not wishing to use an XOF may use `hash_to_field` defined in Section 5.3 of [@!I-D.irtf-cfrg-hash-to-curve], combined with `expand_message_xmd` defined in Section 5.4.1 of the same document, in place of `hash_to_scalar`. In that case, every element outputted by `hash_to_field` that is equal to 0 MUST be rejected. If that occurs, one should calculate more field elements (using `hash_to_field`), until they get `n` non-zero elements (for example, if there is only one 0 in the output of `hash_to_field(msg, 2)` one must try to calculate `hash_to_field(msg, 3)` etc.).
```
tplooker, line 650: I assume whether you are using an XOF vs a conventional HASH alg is a decision for the ciphersuite, how is that captured?

BasileiosKal: Thanks! That's a good point! Added an additional option in the ciphersuites: `hashing_elements_to_scalars`, to capture the two options.

My thinking: When it comes to hash-to-scalar, i don't think it needs to be captured in the ciphersuite, since it will be used by the sign and spkGen operations, i.e., no-need to communicate the use of HASH vs XOF to a verifying/relying party.

For MapMessageToScallarAsHash (that uses hash-to-scalar), again it may not be necessary (since it's use should be agreed in a higher level at some extent e.g., to which messages it applies) but better be careful.

tplooker (approved):

tplooker:
Same here @BasileiosKal can you resolve conflicts?

BasileiosKal:
> Same here @BasileiosKal can you resolve conflicts?

Yeap. Done! Thanks!

andrewwhitehead (approved):
Looks good, #116 will need to be updated if this is merged.

tplooker:
Multiple approvals, discussed on WG call, open 3 weeks, merging

PR #105, opened 2022-04-19
Clarify keygen optional

### LINKED ISSUE #98: Clarify that the key generation operation is optional
Currently the specification offers a definition for [key generation](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-keygen) however implementations do not have to follow this procedure to be interoperable and instead can design their own, the spec should clarify this.

tplooker:
Resolves #98 plus some minor editorial changes


draft-bbs-signatures.md
```diff
@@ -237,9 +237,13 @@ Throughout the operations of this signature scheme, each message that is signed
 
 The KeyGen algorithm generates a secret key SK deterministically from a secret octet string IKM.
 
+**Note** Implementing this operation as defined below for a particular ciphersuite is RECOMMENDED,
+however implementers are able to define an alternative key generation procedure provided it outputs
+a secret key of the equivalent form.
```
andrewwhitehead, line 242: Maybe we can tighten up 'of the equivalent form'. I think the main requirement is that it chooses keys uniformly from the key space (the scalar field).

BasileiosKal: Agreed. Maybe "...provided that the secret key is statistically close to uniformly random" (taken from the [bls_spec](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.3)) or "...provided that the secret key is infeasible to guess"?? I think the second one covers most security requirements (in the first wording one may wonder how close to uniform is enough, although maybe I'm overthinking it)??

tplooker:
Ready for re-review @andrewwhitehead @BasileiosKal (GH wont let me request re-review)

BasileiosKal (approved):
LGTM

draft-bbs-signatures.md
```diff
@@ -694,15 +694,15 @@ The ZKP protocols use nonces which MUST be different in each context.
 
 ## Choice of underlying curve
 
-BBS signatures can be implemented on any pairing-friendly curve. However care MUST be taken when selecting one that is appropriate, this specification defines a profile for using the BLS12-381 curve in (#ciphersuites) which as a curve currently achieves close to 128-bit security.
+BBS signatures can be implemented on any pairing-friendly curve. However care MUST be taken when selecting one that is appropriate, this specification defines a ciphersuite for using the BLS12-381 curve in (#ciphersuites) which as a curve currently achieves close to 128-bit security.
```
andrewwhitehead, line 697: We could reference http://www.csl.sri.com/papers/vcdm-did-crypto-recs/crypto-review-and-recs-for-VCDM-and-DIDs-implems-FINAL-20211015.pdf here, estimating the strength at 117-120 bits. It's definitely less than 128 bits, but still "virtually one million times stronger than BN-254"

tplooker: Thanks have updated, what do you think?

andrewwhitehead (approved):

draft-bbs-signatures.md
andrewwhitehead: Good enough for me :)

tplooker:
Minor editorial changes, multiple approvals, issue discussed on prior WG call, merging.

PR #108, opened 2022-04-21
Add generator point in G1 in BlindSign

### LINKED ISSUE #107: Add generator in G1 in BlindSign
From the BlindSign function:
https://github.com/decentralized-identity/bbs-signature/blob/506d70c0d266a3b04cb8e8a58796008d5284fd2e/draft-blind-bbs-signatures.md?plain=1#L175

We should add P1, the generator point in G1, in order to make a valid signature after the unblinding:
`
b = commitment + P1 + h0 * s'' + h[j1] * msg[1] + ... + h[jK] * msg[K]
`

If you agree with me I will open a pull request where I will change `b` as proposed and add P1's definition into the Terminology section.

alessandroguggino:
Updated the Blind BBS Signature Extension draft.
Fixes #107 + little updates for consistency to core draft, still many to do.


draft-blind-bbs-signatures.md
```diff
@@ -74,6 +68,16 @@ s'
 blind\_signature
 : The blind digital signature output.
 
+## Notation
+
+Terms specific to pairing-friendly elliptic curves that are relevant to this document are restated in BBS Signature Scheme [TODO add reference], originally defined in [@!I-D.irtf-cfrg-pairing-friendly-curves]
```
tplooker, line 73: Can we replace the TODO and add a reference to the core draft using this link (https://identity.foundation/bbs-signature/draft-bbs-signatures.html) using the reference syntax shown at the bottom here https://raw.githubusercontent.com/decentralized-identity/bbs-signature/main/draft-bbs-signatures.md

tplooker:
Thanks for the PR @alessandroguggino!

draft-blind-bbs-signatures.md
alessandroguggino: Ok, I am going to commit it

alessandroguggino:
Added the reference to the core draft as suggested by @tplooker. Thanks!

tplooker (approved):
LGTM

tplooker:
Purely editorial changes to update the blind sign draft, merging.

PR #112, opened 2022-04-23
More editorial tweaks

tplooker:
- Minor cleanup of language / syntax
- Fixed some broken references
- Shifted the algorithmic explanation to the appendix

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -666,7 +666,7 @@ The IKM input to KeyGen MUST be infeasible to guess and MUST be kept secret. One
 
 Secret keys MAY be generated using other methods; in this case they MUST be infeasible to guess and MUST be indistinguishable from uniformly random modulo q.
 
-BBS signatures are nondeterministic, meaning care must be taken against attacks arising from signing with bad randomness, for example, the nonce reuse attack on ECDSA [@HDWH12]. It is RECOMMENDED that the nonces and presentation messages used in this specification are chosen at random from a trusted source of randomness (see "Presentation message selection" section below for additional considerations).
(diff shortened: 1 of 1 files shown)


tplooker:
- Removed the use of context security consideration because it is addressed in other considerations and context is an undefined term in the spec
- tweaked the language around the recommendations for KeyValidation.

draft-bbs-signatures.md
```diff
@@ -700,7 +666,7 @@ The IKM input to KeyGen MUST be infeasible to guess and MUST be kept secret. One
 
 Secret keys MAY be generated using other methods; in this case they MUST be infeasible to guess and MUST be indistinguishable from uniformly random modulo q.
 
-BBS signatures are nondeterministic, meaning care must be taken against attacks arising from signing with bad randomness, for example, the nonce reuse attack on ECDSA [HDWH12]. It is RECOMMEDNED that the nonces and presentation messages used in this specification are chosen at random from a trusted source of randomness (see "Presentation message selection" section below for additional considerations).
+BBS signatures are nondeterministic, meaning care must be taken against attacks arising from signing with bad randomness, for example, the nonce reuse attack on ECDSA [@HDWH12]. It is RECOMMENDED that the nonces and presentation messages used in this specification are chosen at random from a trusted source of randomness, see the (#presentation-message-selection) for additional considerations.
```
BasileiosKal, line 669:
```suggestion
BBS proofs (SPK's) are nondeterministic, meaning care must be taken against attacks arising from using bad randomness, for example, the nonce reuse attack on ECDSA [@HDWH12]. It is RECOMMENDED that the presentation messages used in this specification contain a nonce chosen at random from a trusted source of randomness, see the (#presentation-message-selection) for additional considerations.
```

This may be a little confusing. Signatures as described in the spec are (or can be) deterministic. Also, the presentation message appears in spkGen and not in sign. For that i suggest making the consideration about the BBS+ proofs instead of the signatures.

The term nonce is also not used in the spec, but the intent is for the nonce to be contained in the presentation message. I think the suggested change reflects that.

andrewwhitehead (approved):

tplooker:
Discussed on 2nd May WG call, resolved to merge, multiple approvals

PR #115, opened 2022-05-02
Update KeyValidate operation

### LINKED ISSUE #109: Update KeyValidate operation definition
With the changes to how we handle the message generators the procedure as defined in this section doesn't make sense now as they are no longer apart of the public key, therefore this should be updated to reflect that.

tplooker:
Fixes #109

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -379,11 +379,11 @@ Procedure:
 
 2. if W is INVALID, return INVALID
 
-2. if subgroup_check(A) is INVALID, return INVALID
+3. if subgroup_check(A) is INVALID, return INVALID
 
-3. If W == Identity_G2, return INVALID
+4. If W == Identity_G2, return INVALID
 
-4. return VALID
+5. return VALID
 ```
 
 ### Sign
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -375,15 +375,11 @@ Outputs:
 
 Procedure:
 
-1. (W, H0, H) = octets_to_point(PK)
+1. W = octets_to_point(PK)
 
 2. If W == Identity_G2, return INVALID
 
-3. result = subgroup_check(W) && subgroup_check(H0)
-
-4. for i in 0 to len(H): result &= subgroup_check(H[i])
-
-5. return result
+3. return subgroup_check(W)
 ```
 
 ### Sign
@@ -416,7 +412,7 @@ Outputs:
 
 Procedure:
 
-1. (W, H0, H) = octets_to_point(PK)
+1. W = octets_to_point(PK)
 
 2. generators =  (H_s || H_d || H_1 || ... || H_L)
 
@@ -471,11 +467,11 @@ Procedure:
 
 1. (A, e, s) = (octets_to_point(signature.A), OS2IP(signature.e), OS2IP(signature.s))
 
-2. pub_key = octets_to_point(PK)
+2. W = octets_to_point(PK)
 
-3. if subgroup_check(A) is INVALID
+3. if subgroup_check(A) is INVALID, return INVALID
 
-4. if KeyValidate(pub_key) is INVALID
+4. if KeyValidate(W) is INVALID, return INVALID
 
 5. generators =  (H_s || H_d || H_1 || ... || H_L)
 
@@ -518,7 +514,7 @@ Parameters:
 
 Outputs:
 
-- proof, octet string.
+- proof, octet string; or INVALID.
 
 Procedure:
 
@@ -528,9 +524,9 @@ Procedure:
 
 3. (j1, j2,..., jU) = [L] \ RevealedIndexes
 
-4. if subgroup_check(A) is INVALID abort
+4. if subgroup_check(A) is INVALID, return INVALID
 
-5. if KeyValidate(PK) is INVALID abort
+5. if KeyValidate(PK) is INVALID, return INVALID
 
 6. generators =  (H_s || H_d || H_1 || ... || H_L)
 
@@ -605,7 +601,7 @@ Outputs:
 
 Procedure:
 
-1. if KeyValidate(PK) is INVALID
+1. if KeyValidate(PK) is INVALID, return INVALID
 
 2. (i1, i2, ..., iR) = RevealedIndexes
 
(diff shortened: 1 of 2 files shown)


tplooker:
Best to merge #95 ahead of this PR

tplooker:
@andrewwhitehead @BasileiosKal

BasileiosKal (approved):
As i said [commended](https://github.com/decentralized-identity/bbs-signature/issues/109#issuecomment-1114727062) in issue #109, I would just remove keyValidate from spkGen. That said, not really very strong opiniated about it, so I’m fine with what the WG decides.

draft-bbs-signatures.md
```diff
@@ -443,11 +439,11 @@ Procedure:
...
 1. (A, e, s) = (octets_to_point(signature.A), OS2IP(signature.e), OS2IP(signature.s))
 
-2. pub_key = octets_to_point(PK)
+2. if subgroup_check(A) is INVALID, return INVALID
 
-3. if subgroup_check(A) is INVALID
+3. if KeyValidate(PK) is INVALID, return INVALID
 
-4. if KeyValidate(pub_key) is INVALID
+4. W = octets_to_point(PK)
```
mikelodder7, line 446: We need to add the key validation check to verify. The Point at infinity will allow any signature to verify. `subgroup_check` won't catch this since the point at infinity is in the subgroup. Should I submit a fix to this or do you want to change it?

draft-bbs-signatures.md
```diff
@@ -604,11 +600,11 @@ Procedure:
 
 8. cv = HASH(PK || Abar || A' || D || C1 || C2 || pm)
```
mikelodder7, line 601: I'm noticing this everywhere but curious about the best approach. All points and scalars that will be used in pairing operations should be checked in verify functions for zero and infinity to avoid returning true for invalid signatures when this is the case. How should we add this method check?

draft-bbs-signatures.md
BasileiosKal: I don't think this PR removes KeyValidate from verify??

draft-bbs-signatures.md
```diff
@@ -443,11 +439,11 @@ Procedure:
 
 1. (A, e, s) = (octets_to_point(signature.A), OS2IP(signature.e), OS2IP(signature.s))
 
-2. pub_key = octets_to_point(PK)
+2. if subgroup_check(A) is INVALID, return INVALID
```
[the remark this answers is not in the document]
BasileiosKal: Don't really see how the signature validates with any PK even if e, s are 0. The signature will just become P1* SK. You will still need the correct PK to validate it?? 

That said, strong +1 for adding the checks for e and s to be non 0 in validate.

draft-bbs-signatures.md
BasileiosKal: That's a good point. So that means A, B, A' and Abar then (since W is already checked for 0 and infinity)?? Maybe a security consideration or an addition to the considerations section describing that all elements passed to the pairings must be checked accordingly (similar to the hash inputs lengths)??

draft-bbs-signatures.md
```diff
@@ -369,15 +369,11 @@ Outputs:
...
+1. W = octets_to_point(PK)
 
 2. If W == Identity_G2, return INVALID
 
-3. result = subgroup_check(W) && subgroup_check(H0)
-
-4. for i in 0 to len(H): result &= subgroup_check(H[i])
-
-5. return result
+3. return subgroup_check(W)
```
tplooker, line 376: Swap so we do the subgroup_check before checking the public key does not equal Identity_G2

draft-bbs-signatures.md
mikelodder7: I'm not saying it does. I'm saying we should add it.

draft-bbs-signatures.md
mikelodder7: That's a good start. For SpkVerify this means A', Abar, and D and scalars. For verify this means A and scalars. The public key should always be checked.

draft-bbs-signatures.md
tplooker: But the function already has it?

draft-bbs-signatures.md
tplooker: Addressed

tplooker:
> As i https://github.com/decentralized-identity/bbs-signature/issues/109#issuecomment-1114727062 in issue https://github.com/decentralized-identity/bbs-signature/issues/109, I would just remove keyValidate from spkGen (which is also covered by PR https://github.com/decentralized-identity/bbs-signature/pull/112). That said, not really very strong opinionated about it, so I’m fine with what the WG decides.

Ok shall we also remove the sub-group check for A and then add it as a security consideration that signature verification should be performed prior to SpkGen?

BasileiosKal:
>Ok shall we also remove the sub-group check for A and then add it as a security consideration that signature verification should be performed prior to SpkGen?
 
Yeap that makes sense to me. For now, at least. Something like @andrewwhitehead's proposal in the WG call may change this approach so i think it would be fine to merge the PR as is also. Since the goal was to update KeyValidate, i think this covers it.

draft-bbs-signatures.md
mikelodder7: Should key validate check for infinity and do the subgroup, on curve checks? That would allow a common check. Or maybe we add a function called `IsValidPoint` and use that to perform all the checks and reuse it where necessary

draft-bbs-signatures.md
```diff
@@ -604,11 +602,11 @@ Procedure:
 
 8. cv = HASH(PK || Abar || A' || D || C1 || C2 || pm)
 
-9. if c != cv return INVALID
+9. if c != cv, return INVALID
 
-10. if A' == 1 return INVALID
+10. if A' == 1, return INVALID
```
mikelodder7, line 607: Need to check Abar != 1

BasileiosKal: Can you clarify why this check is needed?? If it's to avoid passing 1 in the parings i suggest even removing the A' != 1 check and have a more unified way of performing this check for all elements.

mikelodder7: Yes that’s why. I agree to the unified method

draft-bbs-signatures.md
tplooker: Have raised #126 to capture this

draft-bbs-signatures.md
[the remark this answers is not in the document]
tplooker: Have raised #127 to capture this

draft-bbs-signatures.md
tplooker: Referenced in #126

draft-bbs-signatures.md
tplooker: Referenced in #126

andrewwhitehead (approved):

PR #116, opened 2022-05-02
Parameterize the drawn bytes for generator creation

### LINKED ISSUE #94: Parameterise the bytes drawn from the XOF so they are defined by the cipher suite
Throughout the operations defined by the spec there are numerous calls to the XOF where a certain number of bytes are read, via the .read() function. How many bytes are read is dependent on several decisions at the cipher suite level (e.g the curve used) hence the current "magic numbers" like 64 being used in these operations needs to be parameterised in the cipher suite definition.

tplooker:
Fixes #94, note this is a first pass at the PR, please check the wording particularly around the parameters r and k. Also is there ever a case where the bytes drawn from the XOF for generator creation would be different to the bytes drawn to calculate the e and s components of the signature in the [sign operation](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-sign) because if not we could generalize/share this parameter.

Also note I think #101 needs to be merged ahead of this

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -390,7 +390,7 @@ Procedure:
 
 Sign computes a signature from SK, PK, over a vector of messages. This method
 describes deterministic signing. For threshold signing, XOF can be replaced
-with a PRF due to the insecurity of deterministic threshold signing. 
+with a PRF due to the insecurity of deterministic threshold signing.
 
 ```
 signature = Sign(SK, PK, (msg_1,..., msg_L), (H_1,..., H_L), header)
@@ -536,7 +536,7 @@ Procedure:
 
 7. domain = OS2IP(HASH(PK || L || generators || Ciphersuite_ID || header)) mod q
 
-8. for element in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU): 
+8. for element in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU):
 
 9.      element = HASH(PRF(8*ceil(log2(q)))) mod q
 
@@ -663,7 +663,7 @@ Procedure:
 
 4.    while(generator_i == Identity_G1 or generator_i == P1)
 
-5.        candidate = hash_to_curve_g1(h.read(64), dst)
+5.        candidate = hash_to_curve_g1(h.read(generator_no_of_bytes), dst)
 
 6.        if candidate not in generators: generator_i = candidate
 
@@ -718,7 +718,7 @@ Inputs:
 
 Parameters:
 
-- q: non-negative integer. The prime order of the G_1 and G_2 groups, 
+- q: non-negative integer. The prime order of the G_1 and G_2 groups,
      defined by the ciphersuite.
 
 Outputs:
@@ -815,7 +815,10 @@ A cryptographic hash function that takes as an arbitrary octet string input and 
 
 - blind_value_generator_seed: The seed used to calculate the signature blinding value generator (H_s). Similar to the message_generator_seed, there are multiple scopes for the blind_value_generator_seed, with the choices being a global seed, a signer specific seed or a signature specific seed. Also, the ciphersuite MUST define this seed OR how to compute it as a pre-cursor operations to any others.
 
-- signature_dst_generator_seed: The seed for calculating the generator used to sign the signature domain separation tag. The scopes and requirements for this seed are the same as the scopes and requirements of the message_generator_seed and blind_value_generator_seed. 
+- signature_dst_generator_seed: The seed for calculating the generator used to sign the signature domain separation tag. The scopes and requirements for this seed are the same as the scopes and requirements of the message_generator_seed and blind_value_generator_seed.
+
+generator_no_of_bytes
+: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
 
 ## BLS12-381 Ciphersuite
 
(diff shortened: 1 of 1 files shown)


BasileiosKal:
I think that the parameter could be shared. The collision resistance security of shake goes up to a certain level either way, so it will be no advantage drawing a larger number of bytes when creating the generators either way (e.g., more than 64 for shake256 or 32 for shake128) even if you need a lot of generators. In general, you would need 2^256 generators to have around 40% chance of a collision, so the danger is practically non-existent IMO.

draft-bbs-signatures.md
```diff
@@ -761,6 +761,9 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 
 - message_generator_seed: The seed used to generate the message generators which form part of the public parameters used by the BBS signature scheme, Note there are multiple possible scopes for this seed including; a globally shared seed (where the resulting message generators are common across all BBS signatures); a signer specific seed (where the message generators are specific to a signer); signature specific seed (where the message generators are specific per signature). The ciphersuite MUST define this seed OR how to compute it as a pre-cursor operations to any others.
 
+generator_no_of_bytes
+: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
```
BasileiosKal, line 765:
```suggestion
: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(log2(p)+k)/8` for the ciphersuite, where `p` and `k` are parameters from the underlying pairing friendly curve being used.
```

No need to introduce a new parameter (r) IMO.

draft-bbs-signatures.md
```diff
@@ -761,6 +761,9 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 
 - message_generator_seed: The seed used to generate the message generators which form part of the public parameters used by the BBS signature scheme, Note there are multiple possible scopes for this seed including; a globally shared seed (where the resulting message generators are common across all BBS signatures); a signer specific seed (where the message generators are specific to a signer); signature specific seed (where the message generators are specific per signature). The ciphersuite MUST define this seed OR how to compute it as a pre-cursor operations to any others.
 
+generator_no_of_bytes
```
BasileiosKal, line 764:
```suggestion
- generator_no_of_bytes
```

draft-bbs-signatures.md
BasileiosKal:
```suggestion
: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `p` the prime characteristic of the G1 and G2 subgroups and `k` the targeted security level (in bits) of the ciphersuite. 
```
Since we don't describe `p` and `k` anywhere else (they are defined in the ciphersuites we reference, however this must be fixed to be clearer, opened issue #118 for that) added here a suggestion for quick explanation of what they are. This may be nice to have even if #118 is closed, since those parameters will most likely only be defined as references.

draft-bbs-signatures.md
```diff
@@ -761,6 +761,9 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 
 - message_generator_seed: The seed used to generate the message generators which form part of the public parameters used by the BBS signature scheme, Note there are multiple possible scopes for this seed including; a globally shared seed (where the resulting message generators are common across all BBS signatures); a signer specific seed (where the message generators are specific to a signer); signature specific seed (where the message generators are specific per signature). The ciphersuite MUST define this seed OR how to compute it as a pre-cursor operations to any others.
 
+- generator_no_of_bytes
+: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(log2(p)+k)/8` for the ciphersuite, where `p` and `k` are parameters from the underlying pairing friendly curve being used.
+
```
tplooker, line 766:
```suggestion
- generator_no_of_bytes
: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `p` the prime characteristic of the G1 and G2 subgroups and `k` the targeted security level (in bits) of the ciphersuite.
```

draft-bbs-signatures.md
```diff
@@ -761,6 +761,8 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 
 - message_generator_seed: The seed used to generate the message generators which form part of the public parameters used by the BBS signature scheme, Note there are multiple possible scopes for this seed including; a globally shared seed (where the resulting message generators are common across all BBS signatures); a signer specific seed (where the message generators are specific to a signer); signature specific seed (where the message generators are specific per signature). The ciphersuite MUST define this seed OR how to compute it as a pre-cursor operations to any others.
 
+- generator_no_of_bytes
+: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `p` the prime characteristic of the G1 and G2 subgroups and `k` the targeted security level (in bits) of the ciphersuite.
```
tplooker, line 765:
```suggestion
: Number of bytes to draw from the XOF when creating generators as per the operation documented in (#creategenerators). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `p` the prime characteristic of the G1 and G2 subgroups and `k` the targeted security level (in bits) of the ciphersuite.

```

tplooker:
Discussed on WG call 2nd of May, will update to have a common variable that covers drawing bytes from XOF for e and s generation case

BasileiosKal (approved):
LGTM

mikelodder7 (approved):

tplooker:
Multiple approvals, discussed on previous WG call, merging.

PR #120, opened 2022-05-02
Rename presentation message to presentation header

### LINKED ISSUE #110: presentation header vs presentation message
With the introduction of the header + message terminology in #95 the question for me remains whether we call the data structure currently known as presentation message that or presentation header. To me it makes more sense to call this a header and reserve the term message for the selectively disclosable payloads that are protected by the signature. We then have two types of headers 1) signature header set by the issuer/signer and required to be revealed in all proofs and 2) presentation header set by the holder/prover.

tplooker:
Resolves #110


mikelodder7 (approved):
LGTM. Maybe we add an example in another PR of the header

christianpaquin (approved):
Good to remove the ambiguity around the "message" term.

draft-bbs-signatures.md
```diff
@@ -461,12 +461,12 @@ Procedure:
 
 ### SpkGen
 
-A signature proof of knowledge generating algorithm that creates a zero-knowledge proof of knowledge of a signature while selectively disclosing messages from a signature given a vector of messages, a vector of indices of the revealed messages, the signer's public key, and a presentation message.
+A signature proof of knowledge generating algorithm that creates a zero-knowledge proof of knowledge of a signature while selectively disclosing messages from a signature given a vector of messages, a vector of indices of the revealed messages, the signer's public key, and a presentation header.
```
BasileiosKal, line 464:
```suggestion
A signature proof of knowledge generating algorithm that creates a zero-knowledge proof of knowledge of a signature while selectively disclosing messages from a signature given a vector of messages, a vector of indices of the revealed messages, the signer's public key, and a presentation header (see [Presentation header selection](#presentation-header-selection) for more details).
```

BasileiosKal: Just a minor suggestion

BasileiosKal (approved):
LGTM.

tplooker:
Multiple approvals, discussed on WG call, merging.

PR #123, opened 2022-05-04
Hash usage harmonization

christianpaquin:
Harmonized use of hash function (use `HASH` vs `H`), clarified starting salt value in KeyGen, fixed broken MD links and some typos.


brentzundel (approved):

draft-bbs-signatures.md
```diff
@@ -298,20 +298,20 @@ Parameters:
...
 Definitions:
 
-- HKDF-Extract is as defined in [@!RFC5869], instantiated with hash H.
-- HKDF-Expand is as defined in [@!RFC5869], instantiated with hash H.
+- HKDF-Extract is as defined in [@!RFC5869], instantiated with hash function HASH.
+- HKDF-Expand is as defined in [@!RFC5869], instantiated with hash function HASH.
 - I2OSP and OS2IP are as defined in [@!RFC8017], Section 4.
 - L is the integer given by ceil((3 * ceil(log2(q))) / 16).
-- "BBS-SIG-KEYGEN-SALT-" is an ASCII string comprising 20 octets.
+- INITSALT is the octet string "0x4242532d5349472d4b455947454e2d53414c542d", obtained by taking the UTF-8 representation of the ASCII string "BBS-SIG-KEYGEN-SALT-".
```
tplooker, line 305: I think its great to have to octet string here explicitly however I was also thinking that we could update the encoding of elements to be hashed section to include more generally how string based values should be encoded prior to hashing?

draft-bbs-signatures.md
```diff
@@ -714,7 +714,7 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 
 ## BLS12-381 Ciphersuite
 
-H
+HASH
 : SHAKE-256 as defined in [@!SHA3]
```
tplooker, line 718: Just a note we need to come back here and define how many bytes are drawn because this is an extendable output function

draft-bbs-signatures.md
BasileiosKal, line 305: Thank you @christianpaquin. Just a nit. I wonder if it would be better to "enforce" the utf-8 encoding in the [Encoding of elements to be hashed](https://github.com/decentralized-identity/bbs-signature/blob/main/draft-bbs-signatures.md#encoding-of-elements-to-be-hashed) section?? Maybe by adding the following line,
```
- Strings must be encoded to octets using the UTF-8 representation and zero-extending the result to a length that is a multiple of 8 bits.
```
That way the UTF-8 representation will be used not only for the salt in keyGen but for any string that needs to be hashed (for example a DST or the Ciphersuite_ID).

tplooker: Do you think you could address this in this PR?

BasileiosKal: A sorry. I just saw @tplooker's [comment](https://github.com/decentralized-identity/bbs-signature/pull/123#discussion_r865265607) above with the same proposal. Sorry about that.

tplooker: If so I would update this section as follows.

```suggestion
- INITSALT the string value of "BBS-SIG-KEYGEN-SALT-", for the avoidance of doubt as per the rules outlined in (#encoding-of-elements-to-be-hashed) the octet string based representation past to the `HASH` function is "0x4242532d5349472d4b455947454e2d53414c542d".
```

christianpaquin: I didn't see other usage of ASCII strings in the text; all other references (unless I missed something?) take octet string as input (aka byte arrays, right?), which made me think that encoding for these were application-specific.

christianpaquin: As I commented above, the rest of the spec expects octet strings, making it sound like the encoding was done at the application level. If we want to make these input ASCII strings, then we should indeed specify how to transform them in this section.

tplooker: Good catch IMO I think DST should stay as an octet string as the underlying data could be any information not just a string value but we should update the CipherSuite_ID parameter to be a string and therfore its encoding would follow the rule defined above.

tplooker: Right, good catch, I have noted below I think the CipherSuite_ID input value type should be updated to be of type string and therefore would use this rule

christianpaquin: > If so I would update this section as follows.

If we had the ASCII encoding rules to the hash section (which I'll do), I don't think having the explicit value is necessary in that section (feels repetitive? and a precedent to doubly-explain hash procedure everywhere else..)

tplooker (approved):
LGTM

draft-bbs-signatures.md
```diff
@@ -252,6 +252,7 @@ To avoid ambiguity, each element passed to the HASH or the XOF function, includi
 - Non-negative integers must be encoded using `I2OSP` with an output length of 8 bytes.
 - Scalars must be zero-extended to a fixed length, defined by a particular ciphersuite.
 - Octet strings must be zero-extended to a length that is a multiple of 8 bits. Then, the extended value is encoded directly.
+- ASCII strings must be transformed into octet strings using UTF-8 encoding.
```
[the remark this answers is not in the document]
christianpaquin: Yes they are ([here](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-encoding-of-elements-to-be-))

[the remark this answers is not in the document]
tplooker: Ah my mistakes was looking at an old version.

BasileiosKal (approved):
LG

tplooker:
Multiple approvals, purely editorial, merging

Issue #117, opened 2022-05-02
Consider making "messages" and therefore "message generators" optional to all operation APIs


tplooker:
Now we have introduced the concept of an optional header in #95 I want to consider whether we allow for constructions of BBS signatures which do not make use of the "selective disclosure" feature, instead just use BBS as an ordinary group signature. I believe there are several interesting usecases such as privacy preserving access tokens or hardware attestations that would likely use BBS in this form.

In short my proposal is for both the header and messages (along with message generators) to be optional. Meaning a signer could create a signature that has neither, either or both a header and array of messages.

BasileiosKal:
+1. In this case, the issuer will just sign their public key and i dont really see any issues with that. We can specify that L >= 0 (where L the number of messages that can be selectively disclosed).

tplooker:
One thing to check is in making this change whether there is anything different in the SpkGen or SpkVerify operations?

BasileiosKal:
I don’t think there is. There is nothing special about a signature on just 1 message (the domain which will be just the hash of the signers PK)

tplooker:
One other thing to note is that because the messages and message generators are seperate arguments past to the sign verify etc functions by making these arguments optional we create the case where only one could be supplied which would be invalid, for example just messages and no generators or vice versa.

BasileiosKal:
That’s a good point. I think a check must be added to ensure that the number of messages is the same with the number of generators (regardless of what we decide here). IMO this will also resolve this issue.

andrewwhitehead:
The method could also accept an optional list of (generator, message) pairs.

BasileiosKal:
> The method could also accept an optional list of (generator, message) pairs.

We could but IMO it may be hard to get the values out of that list of tuples to be used in the operation, i.e., if the input is `[(H_1, msg_1), (H_2, msg_2), ...., (H_n, msg_n)]` wouldn't it be a little unclear how we will get the generators out to be passed in the `domain` etc.?? 

Further i like passing a list of messages because it makes the API as similar as possible to standard "known" digital signatures, so people can draw clear lines between them (i.e., say stuff like "it's just a digital signature but instead of 1 message you will pass a list of messages" etc), just to help with adoption.

That said, i do recognize that the pairs approach makes sense and I'm definitely not opposed to it. I would like to see a PR to get a better idea of how it is going to look.

tplooker:
[...]

PR #128, opened 2022-05-05
Adding revealed messages to challenge hash

### LINKED ISSUE #74: Adding the revealed messages to the chalenge to avoid forgery
```
```
The proof value will be,
```
```

```
```

1.	`c’  = c ^ -1 mod q`
2.	`msg_adv  = m^_2 * c'`

```
```

```
```
```
```

BasileiosKal:
Closes #74 

Adding the revealed messages as well as the signature domain value to the challenge hash in ProofGen and ProofVerify.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -505,6 +505,7 @@ Inputs:
 - msg_1,..., msg_L, octet strings. Messages in input to Sign.
 - H_1,..., H_L, points of G1. The generators in input to Sign.
 - RevealedIndexes, vector of unsigned integers. Indexes of revealed messages.
+                   RevealedIndexes MUST be ordered from the lowest to the highest value.
 - signature, octet string in output form from Sign.
 - header, an optional octet string containing context and application specific
           information. If not supplied, it defaults to an empty string.
@@ -591,6 +592,7 @@ Inputs:
 - msg_i1,..., msg_iR, octet strings. The revealed messages in input to ProofGen.
 - H_1,..., H_L, points of G1. The generators in input to Sign.
 - RevealedIndexes, vector of unsigned integers. Indexes of revealed messages.
+                   RevealedIndexes MUST be ordered from the lowest to the highest value.
 - header, an optional octet string containing context and application specific
           information. If not supplied, it defaults to an empty string.
 - ph, octet string
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -505,7 +505,7 @@ Inputs:
 - msg_1,..., msg_L, octet strings. Messages in input to Sign.
 - H_1,..., H_L, points of G1. The generators in input to Sign.
 - RevealedIndexes, vector of unsigned integers. Indexes of revealed messages.
-                   RevealedIndexes MUST be ordered from the lowest to the highest value.
+                   This vector MUST be ordered from the lowest to the highest value.
 - signature, octet string in output form from Sign.
 - header, an optional octet string containing context and application specific
           information. If not supplied, it defaults to an empty string.
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -534,45 +534,47 @@ Procedure:
 
 6. generators =  (H_s || H_d || H_1 || ... || H_L)
 
-7. domain = OS2IP(HASH(PK || L || generators || Ciphersuite_ID || header)) mod q
+7. revealed_msgs = (msg_i1 || msg_i2 || ... || msg_iR)
 
-8. for element in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU):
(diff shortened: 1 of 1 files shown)


tplooker (approved):
LGTM

draft-bbs-signatures.md
```diff
@@ -534,45 +534,47 @@ Procedure:
...
+17. s' = s + r2 * r3
 
-17. C1 = A' * e~ + H_s * r2~
+18. C1 = A' * e~ + H_s * r2~
 
-18. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
+19. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
 
-19. c = HASH(PK || Abar || A' || D || C1 || C2 || ph)
+20. c = HASH(PK || Abar || A' || D || C1 || C2 || domain || revealed_msgs || ph)
```
tplooker, line 563: One minor question do we want to include `R` alongside the actual revealed messages like we include L alongside the generators in the domain?

BasileiosKal:
Couple of updates, 
1. Per @tplooker suggestion the R value (number of revealed messages) is added to the challenge hash.
2. Removed the PK from the challenge hash since it is already part of the domain value.

draft-bbs-signatures.md
BasileiosKal: Done. Thanks!

draft-bbs-signatures.md
```diff
@@ -534,45 +534,47 @@ Procedure:
 
 6. generators =  (H_s || H_d || H_1 || ... || H_L)
 
-7. domain = OS2IP(HASH(PK || L || generators || Ciphersuite_ID || header)) mod q
+7. revealed_msgs = (msg_i1 || msg_i2 || ... || msg_iR)
```
tplooker, line 537: On further thought with ProofGen, where is (msg_i1, .., msg_iR) defined?

tplooker: We need to add a step to this procedure that computes msg_i1 ... msg_iR from msg and RevealedIndexes

draft-bbs-signatures.md
```diff
@@ -505,6 +505,7 @@ Inputs:
 - msg_1,..., msg_L, octet strings. Messages in input to Sign.
 - H_1,..., H_L, points of G1. The generators in input to Sign.
 - RevealedIndexes, vector of unsigned integers. Indexes of revealed messages.
+                   RevealedIndexes MUST be ordered from the lowest to the highest value.
```
tplooker, line 508:
```suggestion
                   This vector MUST be ordered from the lowest to the highest value.
```

draft-bbs-signatures.md
```diff
@@ -591,6 +592,7 @@ Inputs:
 - msg_i1,..., msg_iR, octet strings. The revealed messages in input to ProofGen.
 - H_1,..., H_L, points of G1. The generators in input to Sign.
 - RevealedIndexes, vector of unsigned integers. Indexes of revealed messages.
+                   RevealedIndexes MUST be ordered from the lowest to the highest value.
```
tplooker, line 595:
```suggestion
                   This vector MUST be ordered from the lowest to the highest value.
```

tplooker (changes requested):
Have updated with a comment about a way to make the procedure clearer

draft-bbs-signatures.md
```diff
@@ -534,45 +535,47 @@ Procedure:
...
+17. s' = s + r2 * r3
 
-17. C1 = A' * e~ + H_s * r2~
+18. C1 = A' * e~ + H_s * r2~
 
-18. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
+19. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
 
-19. c = HASH(PK || Abar || A' || D || C1 || C2 || ph)
+20. c = HASH(Abar || A' || D || C1 || C2 || R || revealed_msgs || domain || ph)
```
mikelodder7, line 564: We should add the messages too and not just the indices. We need them both. If that’s already the case it’s not clear to me how that is so given revealed_msgs appears to just include indices

BasileiosKal: Thank you! Currently the PR only includes the revealed messages (note that `revealed_msgs =  (msg_i1 || msg_i2 || ... || msg_iR)` where `msg_i1, ..., msg_iR` the revealed messages in the current notation). Including the revealed indices also makes sense to me but there are proposals to remove the indices notion altogether (see for example issue #117 for some context). For that reason, I made the PR a draft for now, until a proper notation is decided.

tplooker:
Will be addressed after merging of #138

BasileiosKal:
Replaced by #188. Closing.

PR #129, opened 2022-05-06
Editorial pass

tplooker:
- Adds a paragraph to handle when optional elements pash to a HASH or XOF function
- Clarifies which inputs/parameters are required vs optional
- Fixes some un-escaped underscores
- Tweaks some language for each operation description
- Reorders parameters in all the operations so required ones are first


draft-bbs-signatures.md
```diff
@@ -388,80 +390,77 @@ Procedure:
...
 
-- msg_1,..., msg_L, octet strings. Messages to be signed.
-- H_1,..., H_L, points of G1. Generators used to sign the messages.
-- SK, a secret key output from KeyGen.
-- PK, a public key output from SkToPk.
-- header, an optional octet string containing context and application specific
-          information. If not supplied, it defaults to an empty string.
+- SK (REQUIRED), an octet string of the form outputted by the KeyGen operation.
+- header (OPTIONAL), an octet string containing context and application specific information. If not supplied, it defaults to an empty string.
+- msg_1,..., msg_L (OPTIONAL), a vector of octet strings. Application specific messages.
```
BasileiosKal, line 402: Not really a big fun of defining the messages as "Application specific". It's definitely correct, however, it may raise the question of "what other messages are there??"

draft-bbs-signatures.md
```diff
@@ -388,80 +390,77 @@ Procedure:
...
+- H_d (REQUIRED), point of G1. The generator used to sign the signature domain.
 
 Outputs:
 
 - signature, an octet string.
 
 Procedure:
 
-1. (W, H0, H) = octets_to_point(PK)
+1. PK = SkToPk(SK)
```
BasileiosKal, line 417: Although i like the simpler API with not having to supply the PK this may add unnecessary computations?? The PK will most likely be cached. It doesn't really guard against un trusted issuers either (they can just skip this step).

draft-bbs-signatures.md
```diff
@@ -577,27 +575,26 @@ Procedure:
 
 ### ProofVerify
 
-ProofVerify checks if a signature proof of knowledge is valid given the proof, the signer's public key, a vector of revealed messages, a vector with the indices of these revealed messages, and the presentation header used in ProofGen.
+This operation checks that a proof is valid for a header, vector of revealed messages (along side their corresponding originally index position when signed) and presentation header against a public key (PK).
```
BasileiosKal, line 578:
```suggestion
This operation checks that a proof is valid for a header, vector of revealed messages (along side their index corresponding to their original position when signed) and presentation header against a public key (PK).
```
Just a minor suggestion though the original is also completely fine. So really not necessary to use the alternative.

draft-bbs-signatures.md
```diff
@@ -273,11 +273,13 @@ As an example of the above transformations, consider the following. Assume that
 
 For readability, this document makes these transformations implicitly, but they MUST precede every call to the HASH or XOF function.
 
+Optional input/parameters to operations that feature in a call to a HASH or XOF function, that are not supplied to the operation should be omitted from the call, for example if X is an optional input/parameter whilst A and B are required then the procedural step of `HASH(A || X || B)` MUST be evaluated to `HASH(A || B)`.
```
BasileiosKal, line 276:
```suggestion
Optional input/parameters to operations that feature in a call to a HASH or XOF function, that are not supplied to the operation should default to the empty string. For example, if X is an optional, not supplied, input/parameter whilst A and B are required, then the procedural step of `HASH(A || X || B)` MUST be evaluated to `HASH(A || "" || B)`.
```
Since, in some places where we hash optional values we specify that if not supplied they default to the empty string, the above might be more accurate??

BasileiosKal (commented):
I really like the direction of this PR. Just some minor comments.

draft-bbs-signatures.md
tplooker: Im happy to keep supplying the public key to the operation API, provided we add a consideration that implementations SHOULD validate the relationship between SK and PK, does that work as a compromise?

draft-bbs-signatures.md
tplooker: What about empty octet string to be more specific?

BasileiosKal: Yeap. That sounds better.

draft-bbs-signatures.md
BasileiosKal: If we are worried about issuers making mistakes then yes a consideration sounds as a better option. Maybe another option could also be to make the relationship clear in the definition of the inputs in Sign? Something like;

```
- SK (REQUIRED), an octet string of the form outputted by the KeyGen operation.
- PK (REQUIRED), an octet string encoded public key output from SkToPk on input the above SK.
```

draft-bbs-signatures.md
tplooker: Have removed for now, we can further refine this language later.

draft-bbs-signatures.md
```diff
@@ -330,18 +325,18 @@ Procedure:
 8. return SK
 ```
 
-**Note** This operation is the RECOMMENDED way of generating secret keys, but its use is not required for compatibility, and implementations MAY use a different KeyGen procedure.  For security, such an alternative KeyGen procedure MUST output SK that is statistically close to uniformly random in the range 1 <= SK < r.
+**Note** This operation is the RECOMMENDED way of generating a secret key, but its use is not required for compatibility, and implementations MAY use a different KeyGen procedure. For security, such an alternative MUST output a secret key that is statistically close to uniformly random in the range 1 <= SK < r.
```
BasileiosKal, line 328:
```suggestion
**Note** This operation is the RECOMMENDED way of generating a secret key, but its use is not required for compatibility, and implementations MAY use a different KeyGen procedure. For security, such an alternative MUST output a secret key that is statistically close to uniformly random in the range 0 < SK < r.
```
We use the 0 < SK < r noation above i think.

BasileiosKal (commented):
Minor nit

mikelodder7 (commented):
Just a small correction but otherwise LGTM

draft-bbs-signatures.md
```diff
@@ -198,7 +190,7 @@ octets\_to\_point(ostr) -> P
 : returns the point P corresponding to the canonical representation ostr, or INVALID if ostr is not a valid output of point\_to\_octets.  This operation is also known as deserialization.
 
 subgroup\_check(P) -> VALID or INVALID
-: returns VALID when the point P is an element of the subgroup of order p, and INVALID otherwise. This function can always be implemented by checking that p \* P is equal to the identity element.  In some cases, faster checks may also exist, e.g., [Bowe19].
+: returns VALID when the point P is an element of the subgroup of order p, and INVALID otherwise. This function can always be implemented by checking that p \* P is equal to the identity element.  In some cases, faster checks may also exist, e.g., [@Bowe19].
```
tplooker, line 193:
```suggestion
: returns VALID when the point P is an element of the subgroup of order p, and INVALID otherwise. This function can always be implemented by checking that q \* P is equal to the identity element.  In some cases, faster checks may also exist, e.g., [@Bowe19].
```

draft-bbs-signatures.md
```diff
@@ -198,7 +190,7 @@ octets\_to\_point(ostr) -> P
 : returns the point P corresponding to the canonical representation ostr, or INVALID if ostr is not a valid output of point\_to\_octets.  This operation is also known as deserialization.
 
 subgroup\_check(P) -> VALID or INVALID
-: returns VALID when the point P is an element of the subgroup of order p, and INVALID otherwise. This function can always be implemented by checking that p \* P is equal to the identity element.  In some cases, faster checks may also exist, e.g., [Bowe19].
+: returns VALID when the point P is an element of the subgroup of order p, and INVALID otherwise. This function can always be implemented by checking that q \* P is equal to the identity element.  In some cases, faster checks may also exist, e.g., [@Bowe19].
```
tplooker, line 193:
```suggestion
: returns VALID when the point P is an element of the subgroup of order q, and INVALID otherwise. This function can always be implemented by checking that q \* P is equal to the identity element.  In some cases, faster checks may also exist, e.g., [@Bowe19].
```

BasileiosKal (approved):

mikelodder7 (approved):

tplooker:
Multiple approvals, discussed on last WG call, merging.

PR #133, opened 2022-05-09
Signature encoding/decoding rules

### LINKED ISSUE #127: Validate e and s signature components non-zero
Raised by @mikelodder7 [here](https://github.com/decentralized-identity/bbs-signature/pull/115/files#r862931354)

tplooker:
~~A draft PR to document the encoding decoding rules for a signature, very early stage, keen for feedback~~

Adds an encoding and decoding function for signatures, fixes #127

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -774,21 +774,25 @@ Parameters:
 
 Outputs:
 
-- A map comprised of the underlying signature components `A` `e` and `s`, OR INVALID
+- A map comprised of the underlying signature components `A` `e` and `s`, OR the result INVALID
 
 Procedure:
 
-1. if len(signature_octets) != octet_point_length + len(e) + len(s), return INVALID // TODO
+1. if len(signature_octets) != (octet_point_length + 2 * xof_no_of_bytes), return INVALID
 
-2. a_octets = signature_octets[0..octet_point_length - 1]
+2. a_octets = signature_octets[0..(octet_point_length - 1)]
 
 3. A = octets_to_point(a_octets)
 
-Check A in otherways here should I be calling IsValidPoint?
+4. if A is INVALID, return INVALID // TODO Check A in other ways here should we be calling IsValidPoint?
 
-4. if A is INVALID, return INVALID
+5. index = octet_point_length
 
-5.
+5. e = OS2IP(signature_octets[index..(index + xof_no_of_bytes - 1])) mod q
+
+6. index += xof_no_of_bytes
+
+6. s = OS2IP(signature_octets[index..(index + xof_no_of_bytes - 1)]) mod q
 
 6. return (A, e, s)
 ```
@@ -816,12 +820,11 @@ Procedure:
 
 1. a_octets = point_to_octets(A)
 
-2. e_octets = I2OSP(e)
+2. e_octets = I2OSP(e, xof_no_of_bytes)
 
-3. s_octets = I2OSP(s)
+3. s_octets = I2OSP(s, xof_no_of_bytes)
 
-4. return [ ...a_octets, ...e_octets, ...s_octets ]
-// TODO document the spread operation??
+4. return [ ...a_octets, ...e_octets, ...s_octets ] // TODO document the spread operation??
 ```
 
 # Security Considerations
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -157,6 +157,9 @@ a || b
 I \\ J
 : For sets I and J, denotes the difference of the two sets i.e., all the elements of I that do not appear in J, in the same order as they were in I.
 
+X\[a..b\]
+: Denotes a slice of the array `X` containing all elements from index `a` until `b`. Note when this syntax is applied to an octet string, each element in the array `X` is assumed to be a single byte.
+
 Terms specific to pairing-friendly elliptic curves that are relevant to this document are restated below, originally defined in [@!I-D.irtf-cfrg-pairing-friendly-curves]
 
 E1, E2
@@ -434,7 +437,7 @@ Procedure:
 
 11. A = B * (1 / (SK + e))
 
-12. signature = (point_to_octets_min(A), e, s)
+12. signature = signature_to_octets(A, e, s)
 
 13. return signature
 ```
@@ -470,27 +473,29 @@ Outputs:
 
 Procedure:
 
-1. (A, e, s) = (octets_to_point(signature.A), OS2IP(signature.e), OS2IP(signature.s))
+1. signature_result = octets_to_signature(signature)
 
-2. if A is INVALID return INVALID
+2. if signature_result is INVALID, return INVALID
 
-3. if subgroup_check(A) is INVALID, return INVALID
+3. (A, e, s) = signature_result
 
-4. if KeyValidate(PK) is INVALID, return INVALID
+4. pub_key = octets_to_point(PK)
 
-5. W = octets_to_point(PK)
+5. if subgroup_check(A) is INVALID
 
-6. generators =  (H_s || H_d || H_1 || ... || H_L)
+6. if KeyValidate(pub_key) is INVALID
 
-7. domain = OS2IP(hash(PK || L || generators || Ciphersuite_ID || header)) mod q
+7. generators =  (H_s || H_d || H_1 || ... || H_L)
 
-8. B = P1 + H_s * s + H_d * domain + H_1 * msg_1 + ... + H_L * msg_L
+8. domain = OS2IP(HASH(PK || L || generators || Ciphersuite_ID || header)) mod q
 
-9. C1 = e(A, W + P2 * e)
+9. B = P1 + H_s * s + H_d * domain + H_1 * msg_1 + ... + H_L * msg_L
 
-10. C2 = e(B, P2)
+10. C1 = e(A, W + P2 * e)
 
-11. return C1 == C2
+11. C2 = e(B, P2)
+
+12. return C1 == C2
 ```
 
 ### ProofGen
@@ -529,57 +534,61 @@ Outputs:
 
 Procedure:
 
-1. (A, e, s) = (octets_to_point(signature.A), OS2IP(signature.e), OS2IP(signature.s))
+1. signature_result = octets_to_signature(signature)
 
-2. (i1, i2,..., iR) = RevealedIndexes
+2. if signature_result is INVALID, return INVALID
 
-3. (j1, j2,..., jU) = [L] \ RevealedIndexes
+3. (A, e, s) = signature_result
 
-4. if subgroup_check(A) is INVALID, return INVALID
+4. (i1, i2,..., iR) = RevealedIndexes
 
-5. if KeyValidate(PK) is INVALID, return INVALID
+5. (j1, j2,..., jU) = [L] \ RevealedIndexes
 
-6. generators =  (H_s || H_d || H_1 || ... || H_L)
+6. if subgroup_check(A) is INVALID, return INVALID
 
-7. domain = OS2IP(hash(PK || L || generators || Ciphersuite_ID || header)) mod q
+7. if KeyValidate(PK) is INVALID abort
 
-8. for element in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU):
+8. generators =  (H_s || H_d || H_1 || ... || H_L)
 
-9.      element = hash(PRF(8*ceil(log2(q)))) mod q
+9. domain = OS2IP(hash(PK || L || generators || Ciphersuite_ID || header)) mod q
 
-10.      if element = 0, go back to step 7
+10. for element in (r1, r2, e~, r2~, r3~, s~, m~_j1, ..., m~_jU):
 
-11. B = P1 + H_s * s + H_d * domain + H_1 * msg_1 + ... + H_L * msg_L
+11.      element = hash(PRF(8*ceil(log2(q)))) mod q
 
-12. r3 = r1 ^ -1 mod q
+12.      if element = 0, go back to step 7
 
-13. A' = A * r1
+13. B = P1 + H_s * s + H_d * domain + H_1 * msg_1 + ... + H_L * msg_L
 
-14. Abar = A' * (-e) + B * r1
+14. r3 = r1 ^ -1 mod q
 
-15. D = B * r1 + H_s * r2
+15. A' = A * r1
 
-16. s' = s + r2 * r3
+16. Abar = A' * (-e) + B * r1
 
-17. C1 = A' * e~ + H_s * r2~
+17. D = B * r1 + H_s * r2
 
-18. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
+18. s' = s + r2 * r3
 
-19. c = hash(PK || Abar || A' || D || C1 || C2 || ph)
+19. C1 = A' * e~ + H_s * r2~
 
-20. e^ = e~ + c * e
+20. c = hash(PK || Abar || A' || D || C1 || C2 || ph)
 
-21. r2^ = r2~ + c * r2
+21. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
 
-22. r3^ = r3~ + c * r3
+22. e^ = e~ + c * e
 
-23. s^ = s~ + c * s'
+23. r2^ = r2~ + c * r2
 
-24. for j in (j1, j2,..., jU): m^_j = m~_j + c * msg_j
+24. r3^ = r3~ + c * r3
 
-25. proof = ( A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU))
+25. s^ = s~ + c * s'
 
-26. return proof
+26. for j in (j1, j2,..., jU): m^_j = m~_j + c * msg_j
+
+27. proof = (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU))
+
+28. return proof
 ```
 
 ### ProofVerify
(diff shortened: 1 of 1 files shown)


draft-bbs-signatures.md
```diff
@@ -738,6 +748,76 @@ Procedure:
...
+- e,
+- s,
+
+Outputs:
+
+- signature_octets, octet string.
+
+Procedure:
+
+// Assume point is valid e and s non zero?
```
BasileiosKal, line 810: I would suggest that yes. Sinse this will be a helper function it does not need to be able to handle generic inputs. The signature values will be well formed when creating the signature in Sign.

draft-bbs-signatures.md
```diff
@@ -738,6 +748,76 @@ Procedure:
...
+
+1. a_octets = point_to_octets(A)
+
+2. e_octets = I2OSP(e, xof_no_of_bytes)
+
+3. s_octets = I2OSP(s, xof_no_of_bytes)
+
+4. return [ ...a_octets, ...e_octets, ...s_octets ] // TODO document the spread operation??
```
BasileiosKal, line 818: Maybe use concatenation instead of spred??

draft-bbs-signatures.md
```diff
@@ -738,6 +748,76 @@ Procedure:
...
+4. if A is INVALID, return INVALID // TODO Check A in other ways here should we be calling IsValidPoint?
+
+5. index = octet_point_length
+
+5. e = OS2IP(signature_octets[index..(index + xof_no_of_bytes - 1])) mod q
+
+6. index += xof_no_of_bytes
+
+6. s = OS2IP(signature_octets[index..(index + xof_no_of_bytes - 1)]) mod q
+
```
BasileiosKal, line 787: Maybe add a line here like???
```
IsVaidPoint(A) or IsValidScalar(e) or IsValidScalar(s) are INVALID, return INVALID
```
Where IsValidScalar(x) checks if the scalar x is 0 < x < q?? Another option could be to validate the signature after de-serializing??

mikelodder7: If we assume implementations always represent valid signatures after deserialization then this could work by only checking it there. However, we should keep it anyway to prevent side-channel attacks where the signature is modified after deserialization in memory and the check is missing.

draft-bbs-signatures.md
mikelodder7: Yes agree

draft-bbs-signatures.md
```diff
@@ -738,6 +748,76 @@ Procedure:
...
+- octet_point_length, the length in bytes of the octet string output by the point_to_octets function
+
+Outputs:
+
+- A map comprised of the underlying signature components `A` `e` and `s`, OR the result INVALID
+
+Procedure:
+
+1. if len(signature_octets) != (octet_point_length + 2 * xof_no_of_bytes), return INVALID
```
tplooker, line 772: This is incorrect, the compressed size of the scalar needs to be captured and denoted correctly.

draft-bbs-signatures.md
```diff
@@ -738,6 +748,76 @@ Procedure:
...
+
+2. a_octets = signature_octets[0..(octet_point_length - 1)]
+
+3. A = octets_to_point(a_octets)
+
+4. if A is INVALID, return INVALID // TODO Check A in other ways here should we be calling IsValidPoint?
```
tplooker, line 778: Subgroup check, also check whether this covers the infinity check.

draft-bbs-signatures.md
```diff
@@ -738,6 +748,76 @@ Procedure:
...
+
+5. index = octet_point_length
+
+5. e = OS2IP(signature_octets[index..(index + xof_no_of_bytes - 1])) mod q
+
+6. index += xof_no_of_bytes
+
+6. s = OS2IP(signature_octets[index..(index + xof_no_of_bytes - 1)]) mod q
```
tplooker, line 786: Check s is non zero

draft-bbs-signatures.md
```diff
@@ -738,6 +748,76 @@ Procedure:
...
+```
+
+### SignatureToOctets
+
+This operation describes how to encode a signature to an octet string.
+
+```
+signature_octets = signature_to_octets(A, e, s)
+
+Inputs:
```
tplooker, line 798: If we assume A e and s are well formed then we are implicitly assuming well formed inputs vs untrusted inputs on the OctetsToSignature operation

tplooker: Add a note that documents this is the case because only invocation of this operation is via the Sign method which already checks A e and s.

draft-bbs-signatures.md
```diff
@@ -432,9 +435,9 @@ Procedure:
 
 10. B = P1 + H_s * s + H_d * domain + H_1 * msg_1 + ... + H_L * msg_L
 
-11. A = B * (1 / (SK + e))
+11. A = B * (1 / (SK + e)) // Check A or B?
 
-12. signature = (point_to_octets_min(A), e, s)
+12. signature = signature_to_octets(A, e, s)
```
BasileiosKal, line 440:
```suggestion
12. signature_octets = signature_to_octets(A, e, s)
```
Just to through an idea. Maybe use signature_octets?? That way in Verify we will have `signature = octets_to_signature(signature_octets)` which i thing is better?? This means that we will need to pass signature_octets as an input to Verify though. So both have tradeoffs.

draft-bbs-signatures.md
```diff
@@ -748,6 +753,77 @@ Procedure:
...
+*Note* this operation deliberately does not perform the relevant checks on the inputs `A` `e` and `s`
+because its assumed these are done prior to its invocation, e.g as is the case with the Sign operation.
+
+```
+signature_octets = signature_to_octets(A, e, s)
+
+Inputs:
+
+- A (REQUIRED), a valid point // TODO need to qualify this more?
+- e (REQUIRED), a non-negative integer representing a valid scalar value
```
BasileiosKal, line 809:
```suggestion
- e (REQUIRED), a non-negative, non-zero, and less than q integer representing a valid scalar value (0 < e < q).
```

draft-bbs-signatures.md
```diff
@@ -748,6 +753,77 @@ Procedure:
...
+- s (REQUIRED), a non-negative integer representing a valid scalar value
```
BasileiosKal, line 810:
```suggestion
- s (REQUIRED), a non-negative, non-zero, and less than q integer representing a valid scalar value (0 < s < q).
```

draft-bbs-signatures.md
```diff
@@ -748,6 +753,77 @@ Procedure:
...
+
+*Note* this operation deliberately does not perform the relevant checks on the inputs `A` `e` and `s`
+because its assumed these are done prior to its invocation, e.g as is the case with the Sign operation.
+
+```
+signature_octets = signature_to_octets(A, e, s)
+
+Inputs:
+
+- A (REQUIRED), a valid point // TODO need to qualify this more?
```
BasileiosKal, line 808:
```suggestion
- A (REQUIRED), a valid point in the G1 subgroup, different than the point at infinity. // TODO need to qualify this more?
```
I think that the above suggestion covers everything we need to specify for A.

draft-bbs-signatures.md
```diff
@@ -748,6 +753,77 @@ Procedure:
...
+### OctetsToSignature
+
+This operation describes how to decode an octet string, validate it and return the underlying components that make up the signature.
+
+```
+(A, e, s) = octets_to_signature(signature_octets)
+
+Inputs:
+
+- signature_octets (REQUIRED), octet string.
```
BasileiosKal, line 765:
```suggestion
- signature_octets (REQUIRED), octet string, output from signature_to_octets on input a valid BBS+ signature (one created by the Sign operation)
```
Maybe make it more explicit here??

draft-bbs-signatures.md
```diff
@@ -748,6 +757,77 @@ Procedure:
...
+
+Parameters:
+
+- octet_point_length, the length in bytes of the octet string output by the point_to_octets function
```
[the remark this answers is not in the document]
tplooker: Do we need to redefine it when its available at the cipher suite definition level? I think we need to figure out the scoping otherwise there would be loads of parameters in every definition?

draft-bbs-signatures.md
tplooker:
```suggestion
- signature_octets (REQUIRED), octet string of the form output from signature_to_octets operation
```

draft-bbs-signatures.md
tplooker: I've inlined the check for now, however happy to add a seperate operation to encapsulate this instead.

draft-bbs-signatures.md
```diff
@@ -748,6 +753,79 @@ Procedure:
...
+
+Procedure:
+
+1. if len(signature_octets) != (octet_point_length + 2 * octet_scalar_length), return INVALID
+
+2. a_octets = signature_octets[0..(octet_point_length - 1)]
+
+3. A = octets_to_point(a_octets)
+
+4. if A is INVALID, return INVALID
```
tplooker, line 781: Whats the best syntax for checking whether A is infinity or not?

draft-bbs-signatures.md
```diff
@@ -862,10 +944,16 @@ signature\_dst\_generator\_seed
...
+: hash\_to\_scalar.
 
 xof\_no\_of\_bytes
-: 64
+: 64.
+
+octet\_scalar\_length
+: 32, based on the RECOMMENDED approach of `ceil(log2(q)/8)`.
+
+octet\_point\_length
```
tplooker, line 955: Isn't this actually specific to the sub-group?

tplooker: if so we should just update the definition

draft-bbs-signatures.md
mikelodder7: We usually denote it as `A != 1_G1`

draft-bbs-signatures.md
mikelodder7: It’s not that it’s recommended it’s the actual field size. This is curve dependent so we should specify as a system parameter with the curve

draft-bbs-signatures.md
```diff
@@ -529,57 +532,59 @@ Outputs:
...
+17. s' = s + r2 * r3
 
-18. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
+18. C1 = A' * e~ + H_s * r2~
 
-19. c = hash(PK || Abar || A' || D || C1 || C2 || ph)
+19. C2 = D * (-r3~) + H_s * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU
 
-20. e^ = e~ + c * e
+20. c = hash(PK || Abar || A' || D || C1 || C2 || ph)
```
mikelodder7, line 573: I think we’re fixing what goes into the hash in another PR right? If not I’ll create one that does

draft-bbs-signatures.md
BasileiosKal: We may want to leave it as a recommendation for applications that may want to apply some custom optimization?? Although i do agree that most cases will use that method. It is also true that the number itself is only base field depended so IMO the definition is fine as it is.

draft-bbs-signatures.md
tplooker: So define `1_G1` as the point at infinity for the G1 subgroup?

draft-bbs-signatures.md
tplooker: Yeap we are

draft-bbs-signatures.md
tplooker: Using `Identity_G1` instead here as it already exists

mikelodder7: 👍🏻

mikelodder7 (approved):
LGTM

draft-bbs-signatures.md
```diff
@@ -157,6 +157,9 @@ a || b
 I \\ J
 : For sets I and J, denotes the difference of the two sets i.e., all the elements of I that do not appear in J, in the same order as they were in I.
 
+X\[a..b\]
+: Denotes a slice of the array `X` containing all elements from index `a` until `b`. Note when this syntax is applied to an octet string, each element in the array `X` is assumed to be a single byte.
```
andrewwhitehead, line 161: Should probably clarify if the range is inclusive or exclusive

draft-bbs-signatures.md
```diff
@@ -829,10 +909,14 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 
 - xof\_no\_of\_bytes: Number of bytes to draw from the xof when performing operations such as creating generators as per the operation documented in (#creategenerators) or computing the e and s components of the signature generated in (#sign). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
 
+- octet\_scalar\_length: Number of bytes to represent a scalar value encoded as an octet string. It is RECOMMENDED this value be set to `ceil(log2(q)/8)`.
+
+- octet\_point\_length: Number of bytes to represent a point encoded as an octet string outputted by the point_to_octets function. It is RECOMMENDED that this value is set to `ceil(log(p)/8)`.
```
andrewwhitehead, line 914: log or log2?

draft-bbs-signatures.md
```diff
@@ -862,10 +946,16 @@ signature\_dst\_generator\_seed
...
 
 xof\_no\_of\_bytes
-: 64
+: 64.
+
+octet\_scalar\_length
+: 32, based on the RECOMMENDED approach of `ceil(log2(q)/8)`.
+
+octet\_point\_length
+: 48, based on the RECOMMENDED approach of `ceil(log(q)/8)`.
```
andrewwhitehead, line 958: `ceil(log2(p)/8)` I think

draft-bbs-signatures.md
tplooker: @andrewwhitehead good point, please check the updated text

draft-bbs-signatures.md
```diff
@@ -748,6 +753,81 @@ Procedure:
...
+
+3. A = octets_to_point(a_octets)
+
+4. if A is INVALID, return INVALID
+
+5. if A == Identity_G1, return INVALID
+
+5. index = octet_point_length
+
+6. e = OS2IP(signature_octets[index..(index + octet_scalar_length - 1]))
```
andrewwhitehead, line 787: Oops, bracket mismatch

draft-bbs-signatures.md
```diff
@@ -748,6 +753,81 @@ Procedure:
...
+
+7. if e <= 0 OR e >= q, return INVALID
```
andrewwhitehead, line 789: I believe OS2IP always returns a non-negative integer?

draft-bbs-signatures.md
```diff
@@ -748,6 +753,81 @@ Procedure:
...
+(A, e, s) = octets_to_signature(signature_octets)
+
+Inputs:
+
+- signature_octets (REQUIRED), octet string of the form output from signature_to_octets operation.
+
+Outputs:
+
+- A, a valid point in the G1 subgroup which is not equal to the identity point.
+- e, a non-negative integer representing a valid scalar value with the range of 0 < e < q.
```
andrewwhitehead, line 770: Seems a bit redundant?

draft-bbs-signatures.md
BasileiosKal: Yea that's correct. Maybe it should be e != 0 then?

draft-bbs-signatures.md
tplooker: Not sure I understand what you mean here, the description is redundant?

BasileiosKal: IMO although the non-negative part is not really needed i would like to keep it to highlight the fact that e can be un signed integer. But removing it and leaving it as 

```
- e, an integer representing a valid scalar value with the range of 0 < e < q.
```
Is also correct.

draft-bbs-signatures.md
```diff
@@ -748,6 +753,81 @@ Procedure:
...
+
+4. if A is INVALID, return INVALID
+
+5. if A == Identity_G1, return INVALID
+
+5. index = octet_point_length
+
+6. e = OS2IP(signature_octets[index..(index + octet_scalar_length - 1)])
+
+7. if e >= q, return INVALID
```
BasileiosKal, line 789:
```suggestion
7. if e = 0 OR e >= q, return INVALID
```
I think we do need to check if e and s are zero right?? @andrewwhitehead??

draft-bbs-signatures.md
```diff
@@ -829,10 +909,14 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 
 - xof\_no\_of\_bytes: Number of bytes to draw from the xof when performing operations such as creating generators as per the operation documented in (#creategenerators) or computing the e and s components of the signature generated in (#sign). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
 
+- octet\_scalar\_length: Number of bytes to represent a scalar value encoded as an octet string. It is RECOMMENDED this value be set to `ceil(log2(q)/8)`.
```
BasileiosKal, line 912:
```suggestion
- octet\_scalar\_length: Number of bytes to represent a scalar value, in the multiplicative group of integers mod q, encoded as an octet string. It is RECOMMENDED this value be set to `ceil(log2(q)/8)`.
```
Maybe we need to make it a little more explicit here so that the ceil(log2(q)/8) recommendation makes more sense

draft-bbs-signatures.md
andrewwhitehead: `s` could potentially be zero for an unblinded signature if `s' + s'' = 0`. If an issuer wants to use `s = 0` or `e = 0` then they could, but I don't think it's great from a security perspective

tplooker: Addressed for e and s

draft-bbs-signatures.md
tplooker: Thanks, addressed

draft-bbs-signatures.md
BasileiosKal: Although i don't really see any security problems with having either e or s be 0 on the top of my head the paper does define them in the multiplicative group mod q so i would prefer if they where non 0 just for that.

BasileiosKal: The point about blind signatures is also very interesting. If we decide to disallow 0 `s` we may have a problem there (not really saying to change it in the core draft, just that it is something we will have to consider when updating the blind signatures spec).

andrewwhitehead (approved):

BasileiosKal (approved):

tplooker:
Multiple reviews, discussed on last WG call, multiple approvals, merging.

christianpaquin:
Ha, glad this was merged! I had an older HTML spec version opened in my browser, and was about to ask how to serialize the signature :)

PR #135, opened 2022-05-11
Updates the sign procedure

tplooker:
This PR updates the sign procedure to account for the case when `(SK + e) = 0 mod q` which is invalid and therefore requires us to re-compute a new `e` value if it were to occur.

Note @BasileiosKal and I also discussed whether we need to check for `A != Identity_G1` however we believe it is impossible for this case to exist based on how the procedure is documented therefore the additional check is not required.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -440,7 +440,7 @@ Procedure:
 13. return signature_octets
 ```
 
-**Note** When computing step 11 of the above procedure there is a small probability that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evalute to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 11, and or `A != Identity_G1` after step 11 to prevent the production of invalid signatures.
+**Note** When computing step 11 of the above procedure there is an extremely small probability (around `2^(-r)`) that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evalute to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 11, and or `A != Identity_G1` after step 11 to prevent the production of invalid signatures.
 
 ### Verify
 
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -440,6 +440,8 @@ Procedure:
 13. return signature_octets
 ```
 
+**Note** When computing step 11 of the above procedure there is a small probability that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evalute to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 11, and or `A != Identity_G1` after step 11 to prevent the production of invalid signatures.
+
 ### Verify
 
 This operation checks that a signature is valid for a given header and vector of messages against a supplied public key (PK).


andrewwhitehead:
While this is incredibly unlikely to happen, it also doesn't seem like a good solution to just abort the program if it does occur. However I worry a little about the possibility of a side-channel attack in comparing e to the SK.

mikelodder7:
@andrewwhitehead is right. We shouldn’t abort or return to step 10 we should do something else like e - 1 as a conditional move to avoid leaking sk via timing attacks.

tplooker:
To tease this out further I think we agree the following

1) The risk of `(SK + e) = 0 mod q` condition being meet is very low
2) The risk however does exist therefore we do need to account for it by providing some guidance in the procedure to implementers on what to do if the case occurred
3) the risk of a timing attack is not in the computation of `(SK + e) = 0 mod q` it self instead its due to the branch condition existing on the outcome of this computation e.g going back to step 10.

If we see risk in the computation of the `(SK + e) = 0 mod q` check itself (which I dont think we do?), we could instead just check `A != Identity_G1` after step 13 however the branch of going back and re-computing e would still exist therefore the remote possibility of some timing information leaking from a sign operation could still exist?

tplooker:
Another option to remove the branching logic is to pre-compute an alternative value for e and use that if the condition is met, removing the possibility of the timing attack @mikelodder7 I believe this was your suggestion in our conversation?

BasileiosKal:
if we do e - 1 or e + 1 shouldn't we then re-check if the new e is 0 again?? unless we specify SK > 1 it seems that we may end up with a similar problem. 

Just to through another (maybe naive) idea, what if we calculate a random, non-zero scalar `u` (the same way we calculate `e` now) with `0 < u < q`  and then we calculate `e` as `e = (q + u - SK) mod q`?? This way `e + SK = u mod q` that we know it's non zero and the whole operation will be constant time??

tplooker:
Note also just aborting or not performing this check at all (current case), also risks leaking information about the SK so we need to address it regardless, I think @mikelodder7 @BasileiosKal suggestion of pre-computing an alternative value for e we can use when this condition is meet, is the most appropriate solution

mikelodder7:
> Note also just aborting or not performing this check at all (current state of spec), also risks leaking information about the SK so we need to address it regardless, I think @mikelodder7 @BasileiosKal suggestion of pre-computing an alternative value for e we can use when this condition is meet, is the most appropriate solution

That being said, current libraries for ECDSA do NOT check their nonce for secret key incompatibility because the likelihood is so low. The ECDSA spec as written in NIST doesn't even discuss it either. So perhaps we are over engineering this.

andrewwhitehead:
I'm happy to have a small note in the security considerations instead of producing a valid signature in this one unlikely case. I don't think we can just return (A=1, e, s) though because then SK can be easily calculated from e.

BasileiosKal:
Agreed. Isn't the possibility of the attack even being effective the same as an attacker trying random secret keys and finding the correct one?? So even if the attack is feasible i don't thing that the overall security level of the scheme actually declines.

tplooker:
> That being said, current libraries for ECDSA do NOT check their nonce for secret key incompatibility because the likelihood is so low. The ECDSA spec as written in NIST doesn't even discuss it either. So perhaps we are over engineering this.

Interesting does that mean in deterministic variations of ECDSA there is at least one combination of SK + payload that you could just never sign as it would always fail because the nonce is derived from the input deterministically?

mikelodder7:
Yes the standard is RFC6979 which makes ECDSA deterministic.

tplooker:
Discussed on WG call today, we have multiple options here I think we need to firm up more

In general we need to decide whether we want the strategy with sign to be:

1. In all cases even when say `(SK + e) = 0 mod r` that the procedure should still produce a valid signature (e.g therefore compute an alternative value for e)
2. OR abort the sign procedure

We also need to decide, what checks we need

1. Is `(SK + e) = 0 mod r` all we need to check?
2. OR is `A != Identity_G1` a substitute?
3. OR do we need to perform both e.g `A != Identity_G1` AND `(SK + e) = 0 mod r`
4. OR are either of these checks just over complicating the procedure because the likely-hood of this occuring is so statistically improbable and therefore we should just adopt what ECDSA has done and not mention it?
5. OR should we discuss this case in the considerations and or text around the procedure instead of making it an explicit step in the procedure?

BasileiosKal:
For the `(SK+e) = 0 mod r` vs `A = 1_G1` argument i think @andrewwhitehead's point (and please correct me if I'm wrong) was that
```
A = B*(1/(SK+e)) = inverse(B * (SK + e))
```
As a result, `SK + e = 0 mod r => A = 1_G1`. 

I don't think it holds though?? (i.e., i think that `P * (1 / x) != inverse(P * x)`).

What do you think @mikelodder7??? ^

mikelodder7:
It does hold though. B^-0 or T=inverse(sk+e) then B^T all produce zero. Especially in practice. All implementations I know if you call invert on 0 either abort or return 0. My preference is to call out the possibility (ever so small) that this could happen but the odds are so far out we don’t implement it when signing

BasileiosKal:
Ok i see. Thanks! So it is an implementation issue and not something that holds in general. So if we are to add the check i would still prefer to just check for `(SK+e) = 0`.

That said, i agree with @mikelodder7 in not having the check and just adding a sort security consideration. In any operation taking a signature as input we can specify that the signature must be valid then. IMO this is the most readable and at the same time "correct" approach.

mikelodder7:
In verify we should perform checks that the signature values are not zero and the public key is not zero either. Anywhere these are used we should be checking them. Always validate inputs. We can assume values generated during the function from hash methods or PRFs are valid

tplooker:
I have revert the changes to sign procedure and added a note, ready for re-review.

draft-bbs-signatures.md
```diff
@@ -440,6 +440,8 @@ Procedure:
 13. return signature_octets
 ```
 
+**Note** When computing step 11 of the above procedure there is a small probability that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evalute to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 11, and or `A != Identity_G1` after step 11 to prevent the production of invalid signatures.
```
BasileiosKal, line 443:
```suggestion
**Note** When computing step 11 of the above procedure there is a extremely small probability (around 2^(-r)) that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evalute to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 11, and or `A != Identity_G1` after step 11 to prevent the production of invalid signatures.
```

BasileiosKal (approved):
Minor suggestion to drive the point home about how small the probability of getting a 0 actually is. LGTM otherwise

andrewwhitehead (approved):

tplooker:
Multiple reviews and approvals, discussed on multiple WG calls, merging.

tplooker:
Multiple reviews and approvals, discussed on multiple WG calls, merging.

PR #137, opened 2022-05-13
Ciphersuite ID definition

### LINKED ISSUE #118: Ciphersuites format
The ciphersuites definition currently used in the spec needs to be updated.
1. There are terms that must be defined by most likely referencing their definition to some other spec. Examples include the base points of G1 and G2, the prime order p of those groups the security level of the ciphersuite (`k` in PR #116). Are there any other definitions we should include??
2. A specific format for the ID of each ciphersuite must also be defined. Something like,
```
Cipher_ID = "BBS_" || H2C_SUITE_ID || GEN_SC || HASH_OUTPUT_LEN || _"
```
where `GEN_SC` will be used to distinguish between the scopes of the generators seed (i.e., global or issuer specific) and `HASH_OUTPUT_LEN` between the number of bytes drawn for each XOF (since the consensus seems to be that we will parameterize that value)??

tplooker:
Defines the Ciphersuite ID format and supplies a value for the BLS12-381 suite, partially addresses #118

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -802,6 +802,22 @@ This section defines the format for a BBS ciphersuite. It also gives concrete ci
 
 ## Ciphersuite Format
 
+### Ciphersuite ID
+
+The following section defines the format of the unique identifier for the ciphersuite denoted `Ciphersuite_ID`. The REQUIRED format for this string is
+
+```
+  "BBS_" || H2C_SUITE_ID || ADD_INFO
+```
+
+  *  Strings in double quotes are ASCII-encoded literals.
+
+  *  H2C\_SUITE\_ID is the suite ID of the hash-to-curve suite used to define the hash_to_curve function.
+
+  *  ADD\_INFO is an optional string indicating any additional information used to uniquely qualify the ciphersuite. When present this value MUST only contain ASCII characters between 0x21 and 0x7e (inclusive), and MUST end with an underscore (0x5f), other than as the last character the string MUST not contain any other underscores (0x5f).
+
+### Additional Parameters
+
 - hash: a cryptographic hash function.
 
 - point\_to\_octets:
@@ -830,6 +846,9 @@ A cryptographic hash function that takes as an arbitrary octet string input and 
 - xof\_no\_of\_bytes: Number of bytes to draw from the xof when performing operations such as creating generators as per the operation documented in (#creategenerators) or computing the e and s components of the signature generated in (#sign). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
 
 ## BLS12-381 Ciphersuite
+
+Ciphersuite\_ID
+: "BBS\_BLS12381G1\_XOF:SHAKE-256\_SSWU\_RO\_"
 
 hash
 : SHAKE-256 as defined in [@!SHA3]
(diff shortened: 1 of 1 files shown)


draft-bbs-signatures.md
```diff
@@ -802,6 +802,22 @@ This section defines the format for a BBS ciphersuite. It also gives concrete ci
 
 ## Ciphersuite Format
 
+### Ciphersuite ID
+
+The following section defines the format of the unique identifier for the ciphersuite denoted `Ciphersuite_ID`. The REQUIRED format for this string is
+
+```
+  "BBS_" || H2C_SUITE_ID || ADD_INFO
```
BasileiosKal, line 810:
```suggestion
  BBS || "_" || H2C_SUITE_ID || "_" || ADD_INFO || "_"
```
It may be better to define the ID to end with a `_` so we will not have to define that ADD_INFO MUST end with an `_`. Also there is an `_` that is needed after the h2c id i think.

christianpaquin (approved):

draft-bbs-signatures.md
tplooker: I think most of the H2C_SUITE_ID's follow with an `_` already? hence I was just following the pattern established by the BLS spec?

tplooker: See https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-14#section-8.8

draft-bbs-signatures.md
```diff
@@ -802,6 +802,22 @@ This section defines the format for a BBS ciphersuite. It also gives concrete ci
...
+```
+
+  *  Strings in double quotes are ASCII-encoded literals.
+
+  *  H2C\_SUITE\_ID is the suite ID of the hash-to-curve suite used to define the hash_to_curve function.
+
+  *  ADD\_INFO is an optional string indicating any additional information used to uniquely qualify the ciphersuite. When present this value MUST only contain ASCII characters between 0x21 and 0x7e (inclusive), and MUST end with an underscore (0x5f), other than as the last character the string MUST not contain any other underscores (0x5f).
```
tplooker, line 817:
```suggestion
  *  ADD\_INFO is an optional string indicating any additional information used to uniquely qualify the ciphersuite. When present this value MUST only contain ASCII characters between 0x21 and 0x7e (inclusive), and MUST end with an underscore (0x5f), other than the last character the string MUST not contain any other underscores (0x5f).
```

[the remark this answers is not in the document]
tplooker: @christianpaquin includes symbols too, its inline with how bls-signatures has defined it, we can narrow to just alphanumeric if we would rather https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-4.1

draft-bbs-signatures.md
BasileiosKal: quite right! my apologies. So the _ after the H2C_SUITE_ID is not needed. I updated the suggestion.

BasileiosKal (approved):
Minor nit. Otherwise LGTM

draft-bbs-signatures.md
tplooker: What about if ADD_INFO is not supplied though? Then it would be a double underscore would it not?

BasileiosKal: Yea, which could provide a way for a verifier to know when ADD_INFO was not supplied?? I do realize that this is unconventional so Im fine with the current approach also.

tplooker:
Multiple approvals, reviewed on last WG call, merging

PR #140, opened 2022-05-14
Change q to r

BasileiosKal:
Change the notation from `q` to `r` for the G1 and G2 order, for consistency with specs like bls-signatures, pfc etc.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -174,7 +174,7 @@ GT
 e
 : G1 x G2 -> GT: a non-degenerate bilinear map.
 
-q
+r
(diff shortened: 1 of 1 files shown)


tplooker (approved):

christianpaquin (approved):
I'm more used to seeing `q` for scalar arithmetic, but I'm ok to use `r` to be consistent with the BLS spec.

tplooker:
For context, academia often denotes modular arithmetic with `q` however the BLS signatures draft and and pairing friendly drafts use `r`, given this draft references those directly it is more important to be consistent with these rather than academia.

tplooker:

Issue #139, opened 2022-05-13
H_s and H_d not defined in BLS ciphersuite


christianpaquin:
The terminology section says that the `H_s` and `H_d` generators are defined by each ciphersuite, but they are not defined in the BLS12-381 Ciphersuite section.

tplooker:
Agree this is confusing, currently we have `blind_value_generator_seed` and `signature_dst_generator_seed` which are intended to be used with the [CreateGenerators](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-creategenerators) operation to derive H_s and H_d, but this needs to be clarified further

PR #145, opened 2022-05-19
Minor tweak to CreateGenerators

alessandroguggino:
I found the same issue as @christianpaquin in #139, and I hope this helps a little bit to make it clearer.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -653,14 +653,15 @@ Procedure:
 ### CreateGenerators
 
 The CreateGenerators operation defines how to create a set of generators that form a part of the public parameters used by the BBS Signature scheme to accomplish operations such as Sign, Verify, ProofGen and ProofVerify.
+It is used to compute the set of message generators (H), the signature blinding value generator (H_s), and the signature domain separation tag generator (H_d).
 
 ```
-generators = CreateGenerators(dst, message_generator_seed, length);
+generators = CreateGenerators(dst, seed, length);
 
 Inputs:
 
 - dst (REQUIRED), octet string. Domain Separation Tag.
-- message_generator_seed (REQUIRED), octet string.
+- seed (REQUIRED), octet string.
 - length (REQUIRED), unsigned integer. Number of generators to create from the seed and dst.
 
 Outputs:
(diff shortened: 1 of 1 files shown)


christianpaquin:
`message_generator_seed` appears in more places in the spec, so it wouldn't be sufficient to only modify the `CreateGenerators` section. I'd rather keep this name, however, and replace the variable name in `h = xof(seed)`.

tplooker:
Thanks for the PR, I would rather clarify in the ciphersuite definition how H_d and H_s are computed by using this function alongside this change, I will file a seperate PR to do so

mikelodder7 (approved):

alessandroguggino:
> `message_generator_seed` appears in more places in the spec, so it wouldn't be sufficient to only modify the `CreateGenerators` section. 

I think that `message_generator_seed` is a specific seed for message generators H, while `CreateGenerators` is a generic function used to get H_s and H_d too. That's why I changed the input as seed and added a line in the description.

> I would rather clarify in the ciphersuite definition how H_d and H_s are computed

You are right, I should have added them in the list. If you are going to do it, is ok! Otherwise, I'm available.

Thanks!

draft-bbs-signatures.md
```diff
@@ -653,14 +653,15 @@ Procedure:
 ### CreateGenerators
 
 The CreateGenerators operation defines how to create a set of generators that form a part of the public parameters used by the BBS Signature scheme to accomplish operations such as Sign, Verify, ProofGen and ProofVerify.
+It is used to compute the set of message generators (H), the signature blinding value generator (H_s), and the signature domain separation tag generator (H_d).
```
BasileiosKal, line 656:
```suggestion
It is used to compute the set of message generators (`H_1`, `H_2`, etc.,), the signature blinding value generator (`H_s`), and the signature domain generator (`H_d`).
```
I now its confuting but we don't use the generators as part of a list any more for what i know of (in any place that we do is left over by older versions). I think this more accurate in the current notation. Also, we ended up calling the "domain separation tag" just "domain" if I'm not mistaken.

BasileiosKal (approved):
Minor nit. LGTM otherwise! Thanks!

draft-bbs-signatures.md
alessandroguggino: Thank you! Then should we also edit the seed named `signature_dst_generator_seed` in the Ciphersuite section?

BasileiosKal: That's a good point! Made the update in #152. Thanks!

tplooker:
Related to #161

alessandroguggino:
I think I can close this PR after the merge of a new `create_generators`

PR #147, opened 2022-05-19
Update point serialization/de-serialization rules

### LINKED ISSUE #146: Clarify point serialization
The ciphersuite's `octets_to_point` is ambiguous, as it is unclear which point type should be returned (from G1 or G2). This can be inferred by cross-referencing other parts of the spec, but implementers would benefit from a clearer description; e.g., either adding a `_g1` of `_g2` suffix to the method or a type input. The BLS spec support two modes (where G1 and G2 are swapped), so the ambiguity is by design; BBS fixed the group, so we don't need to keep the BLS notation. `point_to_octects` is not as problematic, since presumably you can infer the input type, but we should harmonize both function names.

Another clarification question: both `point_to_octets` and `octets_to_point` refer to the CRFG pairing-friendly curves spec's Appendix C, but doesn't mention which mode to use (compressed or not).  As an example, the BLS spec uses compressed serialization (see its [Appendix A](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#appendix-A))

tplooker:
Fixes #146

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -190,7 +190,7 @@ point\_to\_octets_g1(P) -> ostr, point\_to\_octets_g2(P) -> ostr
 : returns the canonical representation of the point P for the respective subgroup as an octet string. This operation is also known as serialization.
 
 octets\_to\_point_g1(ostr) -> P, octets\_to\_point_g2(ostr) -> P
-: returns the point P for the respective subgroup corresponding to the canonical representation ostr, or INVALID if ostr is not a valid output of the respective point\_to\_octets_G* function.  This operation is also known as deserialization.
+: returns the point P for the respective subgroup corresponding to the canonical representation ostr, or INVALID if ostr is not a valid output of the respective point\_to\_octets_g* function.  This operation is also known as deserialization.
 
 subgroup\_check(P) -> VALID or INVALID
 : returns VALID when the point P is an element of the subgroup of order p, and INVALID otherwise. This function can always be implemented by checking that p \* P is equal to the identity element.  In some cases, faster checks may also exist, e.g., [@Bowe19].
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -928,7 +928,7 @@ point\_to\_octets_g1
 : follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves] for the G1 subgroup, using compression (i.e., setting C\_bit = 1).
 
 point\_to\_octets_g2
-: follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves] for the G2 subgroup.
+: follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves] for the G2 subgroup, using compression (i.e., setting C\_bit = 1).
 
 octets\_to\_point_g1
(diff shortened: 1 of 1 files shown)


tplooker:
@christianpaquin

draft-bbs-signatures.md
```diff
@@ -186,11 +186,11 @@ Identity\_G1, Identity\_G1
...
 
-point\_to\_octets(P) -> ostr
-: returns the canonical representation of the point P as an octet string. This operation is also known as serialization.
+point\_to\_octets_g1(P) -> ostr, point\_to\_octets_g2(P) -> ostr
+: returns the canonical representation of the point P for the respective subgroup as an octet string. This operation is also known as serialization.
 
-octets\_to\_point(ostr) -> P
-: returns the point P corresponding to the canonical representation ostr, or INVALID if ostr is not a valid output of point\_to\_octets.  This operation is also known as deserialization.
+octets\_to\_point_g1(ostr) -> P, octets\_to\_point_g2(ostr) -> P
+: returns the point P for the respective subgroup corresponding to the canonical representation ostr, or INVALID if ostr is not a valid output of the respective point\_to\_octets_G* function.  This operation is also known as deserialization.
```
tplooker, line 193:
```suggestion
: returns the point P for the respective subgroup corresponding to the canonical representation ostr, or INVALID if ostr is not a valid output of the respective point\_to\_octets_g* function.  This operation is also known as deserialization.
```

draft-bbs-signatures.md
```diff
@@ -918,11 +924,17 @@ A cryptographic hash function that takes as an arbitrary octet string input and
 hash
 : SHAKE-256 as defined in [@!SHA3].
 
-point\_to\_octets
-: follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves].
+point\_to\_octets_g1
+: follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves] for the G1 subgroup.
```
BasileiosKal, line 928:
```suggestion
: follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves] for the G1 subgroup, using compression (i.e., setting C\_bit = 1).
```
I think we are using compression in the spec to this point. We can always fall back if need be.

draft-bbs-signatures.md
```diff
@@ -918,11 +924,17 @@ A cryptographic hash function that takes as an arbitrary octet string input and
...
+
+point\_to\_octets_g2
+: follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves] for the G2 subgroup.
```
BasileiosKal, line 931:
```suggestion
: follows the format documented in Appendix C section 1 of [@!I-D.irtf-cfrg-pairing-friendly-curves] for the G2 subgroup, using compression (i.e., setting C\_bit = 1).
```

christianpaquin (approved):

BasileiosKal:
I think that the last step of `SkToPk` should also be updated to `point_to_octets_g2`. First step of `SignatureToOctets` should be `point_to_octets_g1`. Otherwise LGTM.

tplooker:
> I think that the last step of SkToPk should also be updated to point_to_octets_g2. First step of SignatureToOctets should be point_to_octets_g1. Otherwise LGTM.

Good catch, addressed

mikelodder7 (approved):

tplooker:
Multiple approvals, merging.

PR #151, opened 2022-05-21
Alternative hash_to_scalar without an xof

BasileiosKal:
First shot in a `hash_to_scalar_xmd` function to be used with regular hashes as an alternative to the one currently in the spec (that uses an xof). The proposed function essentially makes use of `expand_message_xmd` (not called directly, see below why), and will give a less efficient option for implementers not wishing to make use of shake256.

Comparison with `hash_to_field` and `expand_message_xmd`:
1.	**Memory efficiency**: `hash_to_field` will call `expand_message` once to output the total number of bytes needed for all the outputted elements. Since we may potentially need to create many scalars, this may not be the more efficient approach. For example, if one wants to create a proof with 1000 hidden messages, the output of `expand_message` will need to be at least 64 kb. In contrast, `hash_to_scalar_xmd` will only need to “hold” 64 bytes at once. Note that the total number of hash invocations is the same in both cases, so the computational overhead will effectively be zero.
2.	**Efficiently checking if a scalar is 0**: Although there are discussions on not overengineering the spec, I feel here is the one place we may need to have the check. The reason is that many signed messages may result to the probability of returning a 0 been higher than what we anticipate (not really an issue for bls12-381 but may become an issue for “smaller” ECs).
3.	Calling `expand_message_xmd` directly instead of the xof would be wasteful (it would result to at least n-1 unnecessary hash computations for b_0)

Another option can be to use hkdf instead. The tradeoff here is double the hash computations vs spec simplicity + the wider availability of hkdf.


BasileiosKal:
If the memory usage turns out to not be an issue the operation can be significantly simplified to something like,

```
result = hash_to_scalar_xmd(msg, n)

Procedure:

1. msg_hash= hash(msg || I2OSP(0, 1) || I2OSP(n, 8))

2. (scalar_1, ..., scalar_n) = hash_to_field(msg_hash, n)

3. if 0 in (scalar_1, ..., scalar_n):

4.     msg_hash = hash(msg_hash)

5.     go back to step 2

6. return (scalar_1, ..., scalar_n)
```

draft-bbs-signatures.md
```diff
@@ -715,14 +718,20 @@ Procedure:
 5. return result
 ```
 
-### Hash to scalar
+### Hashing to scalaras
```
tplooker, line 721:
```suggestion
### Hashing to scalar's
```

draft-bbs-signatures.md
```diff
@@ -731,24 +740,85 @@ Inputs:
 
 Parameters:
 
-- r (REQUIRED), non-negative integer. The prime order of the G_1 and G_2 groups,
-     defined by the ciphersuite.
+- r (REQUIRED), non-negative integer. The prime order of the G_1 and G_2 groups, defined by the ciphersuite.
```
tplooker, line 743:
```suggestion
- r (REQUIRED), non-negative integer. The prime order of the G1 and G2 groups, defined by the ciphersuite.
```

draft-bbs-signatures.md
```diff
@@ -730,14 +733,20 @@ Procedure:
...
 ```
 
-### Hash to scalar
+### Hashing to scalars
+
+This operation describes how to hash an arbitrary octet string to `n` scalar values in the multiplicative group of integers mod r (i.e., values in the range [1, r-1]).  This procedure acts as a helper function, and it is used internally in various places within the operations described in the spec. To map a message to a scalar that would be passed as input to the [Sign](#sign), [Verify](#verify), [ProofGen](#proofgen) and [ProofVerify](#proofgen) functions, one must use [MapMessageToScalarAsHash](#mapmessagetoscalar) instead.
+
+This document defines two different hash_to_scalar operations
+1. `hash_to_scalar_xof`, is the more performant option which makes use of an extendable output function, making it ideal for use with functions like SHAKE256 or SHKAE128.
+2. `hash_to_scalar_xmd`, is less performant in comparison, but does not require an extendable output function, making it more suitable for use with hash functions like SHA2 or SHA3.
```
mikelodder7, line 742:
```suggestion
2. `hash_to_scalar_xmd`, is less performant in comparison, requires a fixed length output function like SHA2 or SHA3.
```

BasileiosKal: Thanks! Updated!

draft-bbs-signatures.md
```diff
@@ -746,24 +755,87 @@ Inputs:
...
+Inputs:
+- msg_octets (REQUIRED), octet string. The message to be hashed.
+- n (REQUIRED), non-negative integer. The number of scalars to output.
+
+Parameters:
+
+- r (REQUIRED), non-negative integer. The prime order of the G1 and G2 groups, defined by the ciphersuite.
+- expand_length (REQUIRED), non-negative integer. The number of bytes required to compute each scalar, defined by the ciphersuite.
+- b_in_bytes (REQUIRED), b / 8 for b the output size of H in bits. For example, for b = 256, b_in_bytes = 32.
+- s_in_bytes (REQUIRED), the input block size of H, measured in bytes. For example, for SHA-256, s_in_bytes = 64.
```
andrewwhitehead, line 799: SHA-256 output is 32 bytes

draft-bbs-signatures.md
```diff
@@ -730,14 +733,20 @@ Procedure:
 5. return result
 ```
 
-### Hash to scalar
+### Hashing to scalars
+
+This operation describes how to hash an arbitrary octet string to `n` scalar values in the multiplicative group of integers mod r (i.e., values in the range [1, r-1]).  This procedure acts as a helper function, and it is used internally in various places within the operations described in the spec. To map a message to a scalar that would be passed as input to the [Sign](#sign), [Verify](#verify), [ProofGen](#proofgen) and [ProofVerify](#proofgen) functions, one must use [MapMessageToScalarAsHash](#mapmessagetoscalar) instead.
+
+This document defines two different hash_to_scalar operations
```
andrewwhitehead, line 740: hash_to_scalar needs escaping, or maybe say something like "two different operations for hashing inputs into scalar values"

draft-bbs-signatures.md
```diff
@@ -730,14 +733,20 @@ Procedure:
...
+1. `hash_to_scalar_xof`, is the more performant option which makes use of an extendable output function, making it ideal for use with functions like SHAKE256 or SHAKE128.
```
andrewwhitehead, line 741: I think SHAKE-256 and SHAKE-128 (with the hyphen) for consistency

draft-bbs-signatures.md
```diff
@@ -730,14 +733,20 @@ Procedure:
...
+#### hash_to_scalar_xof
 
-The `hash_to_scalar` procedure hashes elements using an extendable-output function (xof). Applications not wishing to use an xof may use `hash_to_field` defined in Section 5.3 of [@!I-D.irtf-cfrg-hash-to-curve], combined with `expand_message_xmd` defined in Section 5.4.1 of the same document, in place of `hash_to_scalar`. In that case, every element outputted by `hash_to_field` that is equal to 0 MUST be rejected. If that occurs, one should calculate more field elements (using `hash_to_field`), until they get `n` non-zero elements (for example, if there is only one 0 in the output of `hash_to_field(msg, 2)` one must try to calculate `hash_to_field(msg, 3)` etc.).
+The `hash_to_scalar_xof` function takes as an input the message to be hashed and a non-negative integer inticating the number of non-zero scalars to be returned.
```
andrewwhitehead, line 746: *indicating

draft-bbs-signatures.md
```diff
@@ -746,24 +755,87 @@ Inputs:
 
 Parameters:
 
-- r (REQUIRED), non-negative integer. The prime order of the G_1 and G_2 groups,
-     defined by the ciphersuite.
+- r (REQUIRED), non-negative integer. The prime order of the G1 and G2 groups, defined by the ciphersuite.
+- expand_length (REQUIRED), non-negative integer. The number of bytes to read from the xof function, defined by the ciphersuite.
+- dst (REQUIRED), octet string. Domain separation tag.
```
andrewwhitehead, line 760: It seems odd to list cipher suite parameters as required, aren't they all?

draft-bbs-signatures.md
```diff
@@ -746,24 +755,87 @@ Inputs:
...
+
+2. msg_prime = (I2OSP(len(msg), 8) || msg)
+
+3. h = xof(msg_prime || dst_prime || I2OSP(0, 1) || I2OSP(n, 8))
+
+4. for i in (1, ..., n):
+
+5.     scalar_i = OS2IP(h.read(expand_length)) mod r
+
+6.     if scalar_i is 0, go back to step 3
```
andrewwhitehead, line 778: I think it should return to step 5?

draft-bbs-signatures.md
```diff
@@ -746,24 +755,87 @@ Inputs:
...
+#### hash_to_scalar_xmd
+
+Implementations not wishing to use the more performant `hash_to_scalar_xof` operation, can elect to use `hash_to_scalar_xmd`. The `hash_to_scalar_xmd` is based on the `expand_message_xmd` function defined in Section 5.4 of [@!I-D.irtf-cfrg-hash-to-curve], with the addition of checking if the resulting scalar is 0.
+
+```
+result = hash_to_scalar_xmd(msg_octets, n)
+
+Inputs:
+- msg_octets (REQUIRED), octet string. The message to be hashed.
+- n (REQUIRED), non-negative integer. The number of scalars to output.
```
andrewwhitehead, line 792: I think `n` should be greater than 0

draft-bbs-signatures.md
```diff
@@ -746,24 +755,87 @@ Inputs:
...
+
+Parameters:
+
+- r (REQUIRED), non-negative integer. The prime order of the G1 and G2 groups, defined by the ciphersuite.
+- expand_length (REQUIRED), non-negative integer. The number of bytes required to compute each scalar, defined by the ciphersuite.
+- b_in_bytes (REQUIRED), b / 8 for b the output size of H in bits. For example, for b = 256, b_in_bytes = 32.
```
andrewwhitehead, line 798: For consistency maybe just say "the output size of H measured in bytes, equal to b / 8".

Should we just define these during the procedure based on `b` and `s`?

draft-bbs-signatures.md
```diff
@@ -746,24 +755,87 @@ Inputs:
...
+- s_in_bytes (REQUIRED), the input block size of H, measured in bytes. For example, for SHA-256, s_in_bytes = 64.
+- dst (REQUIRED), octet string. Domain separation tag.
+
+Procedure:
+
+1. ell = ceil(expand_length / b_in_bytes)
+
+2. dst_prime = dst || I2OSP(len(dst), 1)
```
andrewwhitehead, line 806: I think the DST length should be prepended for consistency

draft-bbs-signatures.md
```diff
@@ -916,13 +988,13 @@ The parameters that each ciphersuite needs to define are generally divided into
 
 - hash: a cryptographic hash function.
 
-- xof\_no\_of\_bytes: Number of bytes to draw from the xof when performing operations such as creating generators as per the operation documented in (#creategenerators) or computing the e and s components of the signature generated in (#sign). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
+- expand\_length: Number of bytes to draw from the xof when performing operations such as creating generators as per the operation documented in (#creategenerators) or computing the e and s components of the signature generated in (#sign). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
```
andrewwhitehead, line 991: Also applies to the xmd method. Maybe we can escape `e` and `s` here.

draft-bbs-signatures.md
BasileiosKal: Agreed, however there is the `key_info` parameter that is `OPTIONAL`.  IMO this is the reason that every other parameter is `REQUIRED`. Agreed though that this is not ideal, but it is "necessary evil" IMO.

draft-bbs-signatures.md
BasileiosKal: Updated! Thank you!

draft-bbs-signatures.md
BasileiosKal: It's true that this hurts consistency but there is a security reason for that. See [here](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/issues/236) and [here](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/pull/241) for context.

### TL;TR
 prepending len(DST) here will mean that there are different `msg`, `dst`, `expand_length` combinations that give the same hash output. 
 
### Note 
AFAIK there are 2 solutions to the above.

1. prepend the length of the message as well.
2. append the length of the dst only.

Neither are 100% consistent with the [Encoding of elements to be hashed](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-encoding-of-elements-to-be-) section (for option 1 it may not be as "obvious", but after "appending" the lengths of some of the to-be-hashed elements one may thing that we would not need to append the length of the "entire" concatenated result as well).

For that reason, I prefer the current approach (option 2, i also update `hash_to_scalar_xof` to follow this). The "elements to be hashed" section will need updating IMO to something like "concatenating elements" or "Inputs to hash_to_scalar encoding".

Interested to hear others thoughts here as well.

draft-bbs-signatures.md
BasileiosKal: Agreed. Updated to use `b` and `s` instead. Thanks!

draft-bbs-signatures.md
```diff
@@ -746,24 +755,87 @@ Inputs:
...
+5.     scalar_i = OS2IP(h.read(expand_length)) mod r
+
+6.     if scalar_i is 0, go back to step 3
+
+7. return (scalar_1, ..., scalar_n)
+```
+
+#### hash_to_scalar_xmd
+
+Implementations not wishing to use the more performant `hash_to_scalar_xof` operation, can elect to use `hash_to_scalar_xmd`. The `hash_to_scalar_xmd` is based on the `expand_message_xmd` function defined in Section 5.4 of [@!I-D.irtf-cfrg-hash-to-curve], with the addition of checking if the resulting scalar is 0.
```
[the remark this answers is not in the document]
BasileiosKal: Added some text to highlight this fact! Hopefully it is clear enough??

draft-bbs-signatures.md
BasileiosKal: Thank you! Updated the description for `expand_length`.

draft-bbs-signatures.md
BasileiosKal: Thank you! Updated

draft-bbs-signatures.md
tplooker: Im in favour of option 2 also

BasileiosKal:
Updated to use the `range(a, b)` notation introduced to #155.

draft-bbs-signatures.md
```diff
@@ -600,7 +603,7 @@ Procedure:
 
 25. proof = (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU))
 
-26. return proof_to_octets(proof)
+26. return proof
```
tplooker, line 606:
```suggestion
26. return proof_to_octets(proof)
```

andrewwhitehead:
> If the memory usage turns out to not be an issue the operation can be significantly simplified to something like,
> 
> ```
> result = hash_to_scalar_xmd(msg, n)
> 
> Procedure:
> 
> 1. msg_hash= hash(msg || I2OSP(0, 1) || I2OSP(n, 8))
> 
> 2. (scalar_1, ..., scalar_n) = hash_to_field(msg_hash, n)
> 
> 3. if 0 in (scalar_1, ..., scalar_n):
> 
> 4.     msg_hash = hash(msg_hash)
> 
> 5.     go back to step 2
> 
> 6. return (scalar_1, ..., scalar_n)
> ```

Looking again, I think this may be the better option. There's no actual requirement for the hash_to_field implementation to output all of the bytes before generating the scalars, in fact the bls12-381 implementation does it iteratively: https://github.com/zkcrypto/bls12_381/blob/34dab74ae44cd77dfa867928dd2e6c8383d8dc9d/src/hash_to_curve/mod.rs#L50

We could even use the same method for both XOF and XMD modes and eliminate the separate versions, short-circuiting and restarting if a zero scalar is found (which is unlikely anyway) by using a slightly modified version of hash_to_field.

tplooker:
Discussed on WG call, the current sentiment is to restructure the hash to scalar procedure to either use hash_to_field or expand_message to accomplish the same functionality but in a simpler manner. See above for the two proposals from @BasileiosKal and @andrewwhitehead respectively.

tplooker:
Discussed on WG call, resolved to move to the hash_to_field based approach suggested by @BasileiosKal, however a slightly modified procedure to use a counter instead of hash if a zero scalar value is encountered. 

More generally it was also discussed that we should open a tracking issue to determine whether we want to support rejection sampling in general across the spec for procedures that involve generating scalars. There is guidance in the hash to curve draft that advises against this style of check (e.g checking that a generated scalar value is zero) so we may revise to line up with that guidance because the probability is so low.

BasileiosKal:
After some more thought, a potentially annoying issue is that using `hash_to_field` will require defining 2 h2c ciphersuites, one for the xof and one for the xmd version of `expand_message`. The reason is that the h2c spec only defines suites where the `field` in `hash_to_field` is the base field of the EC itself and not the scalars mod r (that we need). This matters because we would then "mod" with the wrong prime (i.e.,  not r).

Any thoughts here?? @andrewwhitehead, @tplooker?? If not, I may move with [Andrew’s proposal](https://github.com/decentralized-identity/bbs-signature/pull/151#issuecomment-1158367638) to avoid this issue.

andrewwhitehead:
@BasileiosKal  The H2C spec does say this..

> The hash_to_field function is also suitable for securely hashing to scalars.  For example, when hashing to the scalar field for an elliptic curve (sub)group with prime order r, it suffices to instantiate hash_to_field with target field GF(r).

However, I'm not sure if we can guarantee that a particular H2C implementation has support for this operation.

tplooker:
> After some more thought, a potentially annoying issue is that using hash_to_field will require defining 2 h2c ciphersuites, one for the xof and one for the xmd version of expand_message. The reason is that the h2c spec only defines suites where the field in hash_to_field is the base field of the EC itself and not the scalars mod r (that we need). This matters because we would then "mod" with the wrong prime (i.e., not r).

On the flip side would this not make the BBS signature ciphersuite definition simpler e.g just a single reference to h2c ciphersuite for all underlying operations?

BasileiosKal:
> However, I'm not sure if we can guarantee that a particular H2C implementation has support for this operation.

Yea those are my worries as well. We can define the "correct" field for `hash-to-field` but there are some implementations that don't support Fr (like [blst](https://github.com/supranational/blst/blob/master/src/hash_to_field.c#L150). Other libs, like [nodle-bls12-381](https://github.com/paulmillr/noble-bls12-381/blob/main/index.ts#L286) and the[ rust crate](https://github.com/zkcrypto/bls12_381/blob/main/src/hash_to_curve/map_scalar.rs) do support this though).

> On the flip side would this not make the BBS signature ciphersuite definition simpler e.g just a single reference to h2c ciphersuite for all underlying operations?

That’s definitely true! However, as Andrew noted, an issue is that not all libraries have support for the operations we need.

Another issue is that we can't just define a different hash-to-field that mods r because we need the mod p to create the generators (i.e., for hash-to-curve). So this can be confusing to implementers (essentially we need two different versions of hash-to-field, one that mods p for create-generators and one that mods r for hash-to-scalar).

BasileiosKal:
Ok! i gave a shot, based on @andrewwhitehead's [proposal](https://github.com/decentralized-identity/bbs-signature/pull/151#issuecomment-1158367638) of using  `expand-message`. As most libraries have a `mod r` functionality, this would create the lowest implementation barrier + keep the spec simpler. I added a note saying that hash-to-scalar essentially calls hash-to-field for Fr, so if implementers have this available, they can just use that. Also made some updates in the ciphersuites. Thoughts??

draft-bbs-signatures.md
```diff
@@ -184,6 +184,9 @@ X\[a..b\]
 range(a, b)
 : For integers a and b, with a <= b, denotes the ascending ordered list of all integers between a and b inclusive (i.e., the integers "i" such that a <= i <= b).
 
+strxor(octet_str_1, octet_str_2)
```
andrewwhitehead, line 187: This isn't used anymore

BasileiosKal: Good catch! Removed!

draft-bbs-signatures.md
```diff
@@ -1121,13 +1147,13 @@ The following ciphersuite is based on the BLS12-381 elliptic curve defined in Se
 
 - hash: SHAKE-256 as defined in [@!SHA3].
 
-- xof\_no\_of\_bytes: 64.
+- expand\_length: 64.
```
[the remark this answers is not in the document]
BasileiosKal: Yea good idea! Done!

draft-bbs-signatures.md
```diff
@@ -662,18 +663,23 @@ Parameters:
...
-- expand_message is the corresponding operation associated with the ciphersuite's hash_to_curve_g1 parameter.
-- seed_dst is the octet string representing "BBS-SIG-GENERATOR-SEED" in the ASCII character encoding.
-- hash_output_bytes is the integer given by ceil((log2(r) + k) / 8).
-- generator_dst is the octet string representing "BBS-SIG-GENERATOR-DST" in the ASCII character encoding.
+- expand_message, as defined by the hash-to-curve suite uniquely
+                  identified by the hash_to_curve_id ciphersuite
+                  parameter.
+- seed_dst, is the octet string representing "BBS-SIG-GENERATOR-SEED"
+            in the ASCII character encoding.
+- expand_length, non-negative integer. Defined by the ciphersuite.
```
andrewwhitehead, line 671: I think this should be in the Parameters section

draft-bbs-signatures.md
```diff
@@ -729,39 +735,60 @@ Procedure:
...
+                    scalars to output.
 
 Parameters:
 
-- r (REQUIRED), non-negative integer. The prime order of the G_1 and G_2 groups,
-     defined by the ciphersuite.
+- expand_message, as defined by the hash-to-curve suite uniquely
+                  identified by the hash_to_curve_id ciphersuite
+                  parameter.
+- dst, octet string. the separation tag defined by the
```
andrewwhitehead, line 756: I'm not sure what DST value this is?

draft-bbs-signatures.md
```diff
@@ -729,39 +735,60 @@ Procedure:
...
 
-4.     if scalar_i is 0, go back to step 3
+4. msg_prime = msg_octets || I2OSP(t, 1) || I2OSP(count, 4)
 
-5. return (scalar_1, ..., scalar_n)
+5. uniform_bytes = expand_message(msg_prime, dst_prime, len_in_bytes)
+
+6. for i in (1, ..., count):
+
+7.     tv = uniform_bytes[(i-1)*expand_length..i*expand_length]
```
andrewwhitehead, line 779:
```suggestion
7.     tv = uniform_bytes[(i-1)*expand_length..i*expand_length-1]
```

draft-bbs-signatures.md
```diff
@@ -1082,13 +1109,13 @@ The parameters that each ciphersuite needs to define are generally divided into
 
 - hash: a cryptographic hash function.
 
-- xof\_no\_of\_bytes: Number of bytes to draw from the xof when performing operations such as creating generators as per the operation documented in (#generator-point-computation) or computing the e and s components of the signature generated in (#sign). It is RECOMMENDED this value be set to one greater than `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
+- expand\_length: Number of bytes to expand a message that will be hashed to a scalar (see [Hashing to Scalars](#hashing-to-scalars)) or to a point in G1 (i.e., a generator, see [Create Generators](#creategenerators)). It is RECOMMENDED that this value is set to one greater or equal to `ceil(r+k)/8` for the ciphersuite, where `r` and `k` are parameters from the underlying pairing friendly curve being used.
```
andrewwhitehead, line 1112: I don't think `ceil(r+k)/8` is correct (it's huge). Maybe `ceil((ceil(log2(p)) + k) / 8)`

draft-bbs-signatures.md
```diff
@@ -729,39 +735,60 @@ Procedure:
...
+                 compute each scalar, defined by the ciphersuite.
 
 Outputs:
 
 - (scalar_1, ..., scalar_n), a list of non-zero scalars mod r.
 
 Procedure:
 
-1. h = xof(msg_octets)
+1. len_in_bytes = cound * expand_length
```
andrewwhitehead, line 767: I realized there's actually an issue with using expand_length here, as it conflicts with the recommended value for hash_to_field which is `ceil((ceil(log2(p)) + k) / 8)`. For BLS12-381 that is 48 bytes, but we have expand_length set to 64 bytes.

draft-bbs-signatures.md
andrewwhitehead: Ah, looks like it's the hash-to-curve ciphersuite identifier. There's a small issue that `hash_to_curve` is defined by the ciphersuite but `hash_to_curve_id` is referenced here and in `create_generators`

draft-bbs-signatures.md
BasileiosKal: Fixed. Thanks!

draft-bbs-signatures.md
BasileiosKal: Updated! Thank you! (named the ciphersuite parameter `hash_to_curve_suite` and used it in `create_generators` and `hash_to_scalar`)

draft-bbs-signatures.md

draft-bbs-signatures.md
BasileiosKal: Good catch! Yea it was wrong. Fixed.

draft-bbs-signatures.md
andrewwhitehead: Yes, 64 is the number of bytes when hashing to a G1 point (32 bytes are used for each component). If we want to allow the use of hash_to_field then I think we would need two different expand_length parameters, or we should just remove it from the ciphersuite and use definitions.

BasileiosKal: Thank you very much for the feedback :) This is clearly confusing so moving to definitions is the best option IMO.

For `ProofGen` I removed the update from here and re-introduce it in PR #183. For `create_generators` and `hash_to_scalar` I made the update here.

draft-bbs-signatures.md
```diff
@@ -1076,19 +1107,17 @@ The following section defines the format of the unique identifier for the cipher
...
 
 - octet\_point\_length: Number of bytes to represent a point encoded as an octet string outputted by the `point_to_octets_g*` function. It is RECOMMENDED that this value is set to `ceil(log2(p)/8)`.
 
-- hashing\_elements\_to\_scalars: either hash_to_scalar using hash (in this case hash MUST be an xof), or hash_to_field with the additional check and re-calculation of more elements until the desired number of non-zero field elements is returned (as described in [Hash to scalar](#hash-to-scalar)).
+- hash\_to\_curve\_suite: The hash-to-curve ciphersuite id, in the form defined in [@!I-D.irtf-cfrg-hash-to-curve], along with a domain separation tag (dst). This defines the hash\_to\_curve\_g1 (the hash\_to\_curve operation for the G1 subgroup, see the [Notation](#notation) section) and the expand\_message (either expand\_message\_xmd or expand\_message\_xof) operations used in this document. Note, the dst defined here only forms the prefix of the actual dst used by hash\_to\_curve\_g1 or expand\_message. The actual value of the dst used by those operations, will be formed by the operation calling them.
```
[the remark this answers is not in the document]
BasileiosKal: Done! (used 2nd option, i.e., using the `hash_to_curve_suite` value to create each dst). Thanks!

draft-bbs-signatures.md
```diff
@@ -1121,13 +1148,13 @@ The following ciphersuite is based on the BLS12-381 elliptic curve defined in Se
 
 - hash: SHAKE-256 as defined in [@!SHA3].
 
-- xof\_no\_of\_bytes: 64.
+- expand\_length: 48.
```
andrewwhitehead, line 1151: I don't think this is used anymore

BasileiosKal: I'm really sorry about that! Fixed. Thanks!

andrewwhitehead (approved):

tplooker (approved):

tplooker:
Multiple reviews and approvals, discussed on multiple wg calls, merging.

PR #152, opened 2022-05-23
Editorial updates

BasileiosKal:
Various editorial changes:
1.	Updated the ascii art to be less than 72 chars (we will need to do that in other places on the spec as well that are inside code blocks etc).
2.	Changed signature_dst_generator_seed -> sig_domain_generator_seed as noted by @alessandroguggino in PR #145 ([comment](https://github.com/decentralized-identity/bbs-signature/pull/145#discussion_r879127346)) 
3.	Fixed wrong reference to the SHA3 specification.
4.	Removed the `hash_to_field` definition from the ciphersuite because it is not used by any operation at the moment. It is only mentioned as an alternative and for that I don’t believe is needed in the ciphersuite. If the proposal from [here](https://github.com/decentralized-identity/bbs-signature/pull/151#issuecomment-1133764513) is adopted will re-add it.
5.	Categorizing the parameters defined on the ciphersuites (there are a lot of parameters and I personally have found my self re-reading the whole section more than once to find the one I was looking for. Maybe this makes things more clear and easy to read??)

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -73,20 +73,43 @@ Below is a basic diagram describing the main entities involved in using the sche
 
 !---
 ~~~ ascii-art
-  (1) sign                                       (3) ProofGen                                        (5) ProofVerify
-  +-----                                         +-----                                               +-----
-  |    |                                         |    |                                               |    |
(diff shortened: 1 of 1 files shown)


tplooker:

PR #138, opened 2022-05-13
API update

BasileiosKal:
Updating the way that the messages and generators are passed to the operations, using @andrewwhitehead's proposal of passing them in pairs. 

The main purpose of the PR is to have a concrete base of discussion about what the best API will be. For alternatives see #117. For some more context see PR #128 (the motivation was “how we get the revealed messages from the (msg_1, …, msg_L) list and the revealed indexes).

IMO all approaches have trade-offs. Although I like the proposal seen in this PR a lot and it makes things more explicit, I wonder if it is more error prone?? That said, it may provide the best balance between being explicit and easy to read. Interested on the WG feedback.

Issue #125, opened 2022-05-05
Refine the relationship/differences between inputs vs parameters for operations


tplooker:
Currently for each operation documented in the spec we have the notion of both inputs and parameters, a convention adopted from the [bls-signature-draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04). 

Loosely what I think we have been following as a largely unwritten rule is that inputs are usually more dynamic or application driven/supplied in nature, where as parameters are more constant across multiple invocations of an operation(s).

To make things more complex we also appear to have two types/scopes of parameters, parameters shared across all operations vs parameters semi-unique to an operation, although it doesn't look like we are following this strictly. 

I think we need to

1) Rationalise the difference between what constitutes an input vs a parameter
2) Decide on the scoping of various parameters and how we notate them in reference to specific operations for example do we inline them all? or reference them from a common section that applies to all operations, or both?

Furthermore I think we need to revisit certain values that may instead be considered parameters instead of inputs, in particular message generators. This is discussed more in #117

BasileiosKal:
One option would be to define that anything that is described by a ciphersuite (i.e., base points etc.,) is a parameter. This will also apply to the message generators.

A stricter definition will be to consider anything that is defined by a ciphersuite AND it is independent from any inputs to be a parameter. This doesn't really apply to the message generators (their number depends to the number of messages).

Regarding 2 my vote is to inline the parameters if we would to consider the message generators as a parameter. If the message generators are supplied as input we can have a separate section to avoid repeated text.

BasileiosKal:
Discussed on the WG on the 6th of February. Will check for any inconsistency and try to close next weak.

BasileiosKal:
Discussed on WG on the 6th of Mar. Will check for consistency and close before publishing draft 02.

tplooker:
Checks were done prior to draft-2 publication, no issue found, closing as a result.

PR #155, opened 2022-05-25
Proof serialization/de-serialization 

### LINKED ISSUE #153: OctetsToProof and ProofToOctets
With the merge of #147 defining the rules for the signature serialisation we now need to define how to do the corresponding for the proof value outputted by the ProofGen operation and consumed by the ProofVerify operation.

BasileiosKal:
Fixes #153

Introduces a `proof_to_octets` and a `octets_to_proof` operation

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -923,7 +923,7 @@ Procedure:
 
 15.     j += 1
 
-16. if index is not equal to length(proof_octets), return INVALID
+16. if index != length(proof_octets), return INVALID
 
 17. Let msg_commitments be an empty set (i.e., msg_commitments = ())
 
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -851,7 +851,7 @@ The `octets_to_proof` operations describes how to parse an octet string represen
 
 The proof value outputted by this operation consists of the following components, in that order:
 
-1. Three (3) valid points of the G1 subgroup, different from the identity point.
+1. Three (3) valid points of the G1 subgroup, each of which must not equal the identity point.
 2. Five (5) integers representing scalars in the range of 1 to r-1 inclusive.
 3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
 
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -952,7 +952,7 @@ Procedure:
 
 1. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU)) = proof
 
-2. Let proof_octets be the empty octet string.
+2. Let proof_octets be an empty octet string.
 
 // Points Serialization.
 3. for point in (A', Abar, D):
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -853,7 +853,7 @@ The proof value outputted by this operation consists of the following components
 
 1. Three (3) valid points of the G1 subgroup, each of which must not equal the identity point.
 2. Five (5) integers representing scalars in the range of 1 to r-1 inclusive.
-3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
+3. A set of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
 
 ```
 proof = octets_to_proof(proof_octets)
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -925,7 +925,7 @@ Procedure:
 
 16. if index != length(proof_octets), return INVALID
 
-17. Let msg_commitments be an empty set (i.e., msg_commitments = ())
+17. msg_commitments = ()
 
 18. If j > 5, set msg_commitments = (s_5, ..., s_(j-1))
 
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -964,7 +964,7 @@ Procedure:
 // Scalar Serialization.
 6. for scalar in (c, e^, r2^, r3^, s^, m^_j1, ..., m^_jU):
 
-7.     scalar_octets = I2OSP(scalar_oct, octet_scalar_length)
+7.     scalar_octets = I2OSP(scalar, octet_scalar_length)
 
 8.     proof_octets = proof_octets || scalar_octets
 
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -926,7 +926,7 @@ This operation describes how to encode a proof, as computed at step 25 in [Proof
 
 The inputed proof value must consist of the following components, in that order:
 
-1. Three (3) valid points of the G1 subgroup, different from the identity point of G1 (i.e., `A', Abar, D`, in ProofGen)
+1. Three (3) valid compressed points of the G1 subgroup, different from the identity point of G1 (i.e., `A', Abar, D`, in ProofGen)
 2. Five (5) integers representing scalars in the range of 1 to r-1 inclusive (i.e., `c, e^, r2^, r3^, s^`, in ProofGen).
 3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof messages (i.e., `m^_j1, ..., m^_jU`, in ProofGen, where U the number of un-revealed messages).
 
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -865,8 +865,8 @@ Inputs:
 
 Parameters:
 
-- r (REQUIRED), non-negative integer. The prime order of the G_1 and
-                G_2 groups, defined by the ciphersuite.
+- r (REQUIRED), non-negative integer. The prime order of the G1 and
+                G2 groups, defined by the ciphersuite.
 - octet_scalar_length (REQUIRED), non-negative integer. The length of
                                   a scalar octet representation, defined
                                   by the ciphersuite.
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -597,7 +597,7 @@ Procedure:
 
 25. proof = (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU))
 
-26. return proof
+26. return proof_to_octets(proof)
 ```
 
 ### ProofVerify
@@ -640,27 +640,31 @@ Procedure:
 
 3. (j1, j2, ..., jU) = [L]\RevealedIndexes
 
-4. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1,...,m^_jU)) = proof
+4. proof_value = octets_to_proof(proof)
 
-5. generators =  (H_s || H_d || H_1 || ... || H_L)
+5. if proof_value is INVALID, return INVALID
 
-6. domain = hash_to_scalar((PK || L || generators || Ciphersuite_ID || header), 1)
+6. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1,...,m^_jU)) = proof_value
 
-7. C1 = (Abar - D) * c + A' * e^ + H_s * r2^
+7. generators =  (H_s || H_d || H_1 || ... || H_L)
 
-8. T = P1 + H_s * domain + H_i1 * msg_i1 + ... H_iR * msg_iR
+8. domain = hash_to_scalar((PK || L || generators || Ciphersuite_ID || header), 1)
 
-9. C2 = T * c + D * (-r3^) + H_s * s^ + H_j1 * m^_j1 + ... + H_jU * m^_jU
+9. C1 = (Abar - D) * c + A' * e^ + H_s * r2^
 
-10. cv = hash_to_scalar((PK || Abar || A' || D || C1 || C2 || ph), 1)
+10. T = P1 + H_s * domain + H_i1 * msg_i1 + ... H_iR * msg_iR
 
-11. if c != cv, return INVALID
+11. C2 = T * c + D * (-r3^) + H_s * s^ + H_j1 * m^_j1 + ... + H_jU * m^_jU
 
-12. if A' == 1, return INVALID
+12. cv = hash_to_scalar((PK || Abar || A' || D || C1 || C2 || ph), 1)
 
-13. if e(A', W) * e(Abar, -P2) != 1, return INVALID
+13. if c != cv, return INVALID
 
-14. return VALID
+14. if A' == 1, return INVALID
+
+15. if e(A', W) * e(Abar, -P2) != 1, return INVALID
+
+16. return VALID
 ```
 
 ### CreateGenerators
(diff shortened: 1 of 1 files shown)


draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+18. return (A_0, A_1, A_2, s_0, s_1, s_2, s_3, s_4, msg_commitments)
+```
+
+### ProofToOctets
+
+This operation describes how to encode a proof, as computed at step 25 in [ProofGen](#proofgen), to an octet string. The input to the operation MUST be a valid proof.
+
+The inputed proof value must consist of the following components, in that order:
+
+1. Three (3) valid points of the G1 subgroup, different from the identity point of G1 (i.e., `A', Abar, D`, in ProofGen)
```
mikelodder7, line 929:
```suggestion
1. Three (3) valid compressed points of the G1 subgroup, different from the identity point of G1 (i.e., `A', Abar, D`, in ProofGen)
```

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+2. Five (5) integers representing scalars in the range of 1 to r-1 inclusive (i.e., `c, e^, r2^, r3^, s^`, in ProofGen).
+3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof messages (i.e., `m^_j1, ..., m^_jU`, in ProofGen, where U the number of un-revealed messages).
```
mikelodder7, line 931: Little endian 32-bit? Can’t remember if we have a description of serialized integers yet

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
 4. return (a_octets || e_octets || s_octets)
 ```
 
+### OctetsToProof
+
+The `octets_to_proof` operations describes how to parse an octet string representation of a proof, validate it and return the underlying components.
```
tplooker, line 850:
```suggestion
This operation describes how to decode an octet string representing a proof, validate it and return the underlying components that make up the proof.
```

Suggestion to make this more consistent language with how we introduce each operation
```suggestion
The `octets_to_proof` operations describes how to parse an octet string representation of a proof, validate it and return the underlying components.
```

tplooker, line 850:
```suggestion
This operation describes how to decode an octet string representing a proof, validate it and return the underlying components that make up the proof.
```

Suggestion to make this more consistent language with how we introduce each operation
```suggestion
The `octets_to_proof` operations describes how to parse an octet string representation of a proof, validate it and return the underlying components.
```

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+
+The proof value outputted by this operation consists of the following components, in that order:
+
+1. Three (3) valid points of the G1 subgroup, different from the identity point.
```
tplooker, line 854:
```suggestion
1. Three (3) valid points of the G1 subgroup, each of which must not equal the identity point.
```

tplooker, line 854:
```suggestion
1. Three (3) valid points of the G1 subgroup, each of which must not equal the identity point.
```

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+2. Five (5) integers representing scalars in the range of 1 to r-1 inclusive.
+3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
```
tplooker, line 856:
```suggestion
3. A set of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
```

Also do we need to clarify the ordering expected here?
```suggestion
3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
```

tplooker, line 856:
```suggestion
3. A set of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
```

Also do we need to clarify the ordering expected here?
```suggestion
3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof message commitments.
```

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+
+Inputs:
+
+- proof_octets (REQUIRED), octet string of the form output from
+                           proof_to_octets operation.
+
+Parameters:
+
+- r (REQUIRED), non-negative integer. The prime order of the G_1 and
+                G_2 groups, defined by the ciphersuite.
```
tplooker, line 869:
```suggestion
- r (REQUIRED), non-negative integer. The prime order of the G1 and
                G2 groups, defined by the ciphersuite.
```
```suggestion
- r (REQUIRED), non-negative integer. The prime order of the G_1 and
                G_2 groups, defined by the ciphersuite.
```

tplooker, line 869:
```suggestion
- r (REQUIRED), non-negative integer. The prime order of the G1 and
                G2 groups, defined by the ciphersuite.
```
```suggestion
- r (REQUIRED), non-negative integer. The prime order of the G_1 and
                G_2 groups, defined by the ciphersuite.
```

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+- octet_scalar_length (REQUIRED), non-negative integer. The length of
```
tplooker, line 870: Just a note we should probably go back through the spec and make this consistent if we want to adopt a pattern of defining parameters that are set by the cipher suite? Shall I create an issue?

tplooker, line 870: Just a note we should probably go back through the spec and make this consistent if we want to adopt a pattern of defining parameters that are set by the cipher suite? Shall I create an issue?

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+Procedure:
+
+1. proof_len_floor = 3 * octet_point_length + 5 * octet_scalar_length
+
+2. if length(proof_octets) < proof_len_floor, return INVALID
+
+// Points (i.e., (A', Abar, D) in ProofGen) de-serialization.
+3. index = 0
+
+4. for i in (0, 2):
```
tplooker, line 890: Just checking this loop syntax is 0,1,2 wonder if we should do `(0,..,2)` and define the syntax in the terminology section?

tplooker, line 890: Just checking this loop syntax is 0,1,2 wonder if we should do `(0,..,2)` and define the syntax in the terminology section?

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+
+13.     if s_j = 0 or if s_j >= r, return INVALID
+
+14.     index += octet_scalar_length
+
+15.     j += 1
+
+16. if index is not equal to length(proof_octets), return INVALID
+
+17. msg_commitments = (s_5, ..., s_(k-1))
```
tplooker, line 918: Where is k defined?

tplooker, line 918: Where is k defined?

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+
+Outputs:
+
+- proof_octets, octet string.
+
+Procedure:
+
+1. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU)) = proof
+
+2. Let proof_octets be the empty octet string.
```
tplooker, line 955:
```suggestion
2. Let proof_octets be an empty octet string.
```
```suggestion
2. Let proof_octets be the empty octet string.
```

tplooker, line 955:
```suggestion
2. Let proof_octets be an empty octet string.
```
```suggestion
2. Let proof_octets be the empty octet string.
```

draft-bbs-signatures.md
```diff
@@ -841,6 +845,132 @@ Procedure:
...
+1. Three (3) valid points of the G1 subgroup, different from the identity point of G1 (i.e., `A', Abar, D`, in ProofGen)
+2. Five (5) integers representing scalars in the range of 1 to r-1 inclusive (i.e., `c, e^, r2^, r3^, s^`, in ProofGen).
+3. A number of integers representing scalars in the range of 1 to r-1 inclusive, corresponding to the un-revealed from the proof messages (i.e., `m^_j1, ..., m^_jU`, in ProofGen, where U the number of un-revealed messages).
+
+```
+proof_octets = proof_to_octets(proof)
+
+Inputs:
+
+- proof (REQUIRED), a BBS proof in the form calculated by ProofGen in
```
tplooker, line 938: Wonder if we actually need to supply the arguments destructed here like we do in signature to octets?

tplooker, line 938: Wonder if we actually need to supply the arguments destructed here like we do in signature to octets?

draft-bbs-signatures.md
BasileiosKal: Thank you! I don't think we discuss this on the document. Aren't there 256 bits necessary for scalars (for bls12-381)?? Also we use I2OSP/OS2IP which operates in big-endian order if I'm not mistaken, so we will need to add a note there if we decide to define integers to be little endian.

draft-bbs-signatures.md
BasileiosKal: Will the points be compressed when calculated in ProofGen?? Wouldn't they be just regular points (not octets outputted by `point_to_octets`)??

draft-bbs-signatures.md
BasileiosKal: Thanks! Updated!

draft-bbs-signatures.md
BasileiosKal: > Also do we need to clarify the ordering expected here?

There is no way that i know of for the ordering to be checked here, so there may be no point in adding such description?? The expected "ordering" is kinda "defined" in step 4 in `ProofVerify` i.e., 

```
4. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1,...,m^_jU)) = proof
```
Since the functions are just utilities to be used internally and there are no inconsistencies that can arise when used with those core operations, an extra definition of the ordering may not be necessary??

draft-bbs-signatures.md
BasileiosKal: Thanks! Updated!

draft-bbs-signatures.md
BasileiosKal: Since the input has no "fixed" length it may be more complicated to define it ''destructed" (i.e., if we define it as `(A', Abar, D, c, e^, r2^, r3^, s^, (m^_1,...,m^_U))` we will need to define `U` which will be hard here or if we define it like `(A', Abar, D, c, e^, r2^, r3^, s^, msg_commitments`) we mix up structured and destructed forms which misses the point of consistency IMO.

In general, for consistency I would prefer passing the signature in a structured way than passing the proof in a destructed one.

draft-bbs-signatures.md
BasileiosKal: Agreed! There is issue #125. We do need to focus more into closing it so the spec will be consistent.

draft-bbs-signatures.md

BasileiosKal: Thank you! Updated

draft-bbs-signatures.md
BasileiosKal: Yeap! Good catch! Updated this line and added the `(a, ..., b)` syntax in the Notation section.

BasileiosKal: Yeap! Done!

draft-bbs-signatures.md
BasileiosKal: Apologies! This was a typo. It should be `j`. Updated.

draft-bbs-signatures.md
```diff
@@ -181,6 +181,9 @@ I \\ J
 X\[a..b\]
 : Denotes a slice of the array `X` containing all elements from and including the value at index `a` until and including the value at index `b`. Note when this syntax is applied to an octet string, each element in the array `X` is assumed to be a single byte.
 
+(a, ..., b)
```
tplooker, line 184:
```suggestion
(a..b)
```

tplooker: OR something like python `range(0,2)`

draft-bbs-signatures.md
tplooker: Happy to move in that direction do we need to raise an issue to capture that?

BasileiosKal: Updated this PR to make the change for `octets_to_signature` and `signature_to_octets` (+ took the opportunity to make the operations < 72 chars). For other operations the update will most likely happen in #138. Opened #163 to keep track of this.

draft-bbs-signatures.md
BasileiosKal: Thanks! Updated to `range(0, 2)` (we use [a..b] for array/octet_string slices)

draft-bbs-signatures.md
```diff
@@ -818,27 +831,156 @@ This operation describes how to encode a signature to an octet string.
...
+
+13.     if s_j = 0 or if s_j >= r, return INVALID
+
+14.     index += octet_scalar_length
+
+15.     j += 1
+
+16. if index is not equal to length(proof_octets), return INVALID
+
+17. Let msg_commitments be an empty set (i.e., msg_commitments = ())
```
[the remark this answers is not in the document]
tplooker: Otherwise from a data representation perspective serialized as JSON you could take the current syntax to mean

```
proof = {
   A_0: ....,
   .....
   msg_commitments: {
      s_5,
      ...
   }
}
```

Instead

```
proof = {
   A_0: ....,
   .....
   s_5,
   ...
}
```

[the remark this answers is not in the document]
BasileiosKal: I think that in the rest of the spec we use it as the first JSON example you listed. For example, in [proofGen](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-proofgen), in step 25 we compute,
```
25. proof = (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU))
```
And in [ProofVerify](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-proofverify), at step 4, we (currently) parse the proof as,
```
4. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1,...,m^_jU)) = proof
```
Show the options I see is
1. We update this on those places as well (will be really easy to do in this PR)
2. We keep using the same notation

I'm fine with both. I don't think it make much difference from an implementations perspective. From an aesthetics perspective i may go with option 1, so there is some distinction between the different parts of the proof, but really not filling strong about it.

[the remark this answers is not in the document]
tplooker: Im actually fine with either so long as we are consistent, suffice to say if you think we are already being consistent w.r.t  option 1 then we can leave it as is.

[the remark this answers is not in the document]
tplooker: There is a pending suggested change attached to this comment also which isn't entirely related to this conversation, once we have a resolution on that I'm happy to approve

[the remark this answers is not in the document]
BasileiosKal: Yeap! Updated based on suggestion. Thanks!

tplooker (approved):

alessandroguggino (approved):

tplooker:
Multiple reviews and approvals, open 13 days, merging

PR #161, opened 2022-05-30
Generator definition within ciphersuite

andrewwhitehead:
This PR moves generator computation into a separate section, allowing for the use of separate computation methods and generator seed options. There are inconsistencies remaining with the test vectors, and the language and security considerations could likely use tidying up.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -256,11 +256,11 @@ In definition of this signature scheme there are two possible variations based u
 
 ### Messages and generators
 
-Throughout the operations of this signature scheme, each message that is signed is paired with a specific generator (point in G1). Specifically, if a generator `H_1` is multiplied with `msg_1` during signing, then `H_1` MUST be multiplied with `msg_1` in all other operations as well (signature verification, proof generation and proof verification). For simplicity, each function will take as input the list of generators to be used with the messages. Those generators can be any distinct element from the generators list `H`. Applications for efficiency can elect to pass the indexes of those generators to the list `H` instead. Care must be taken for the correct generator to be raised to the correct message in that case.
+Throughout the operations of this signature scheme, each message that is signed is paired with a specific generator (point in G1). Specifically, if a generator `H_1` is multiplied with `msg_1` during signing, then `H_1` MUST be multiplied with `msg_1` in all other operations as well (signature verification, proof generation and proof verification). In practice this means that the ordering of the selectively-revealed messages is critical when these are passed as input to this scheme's operations.
 
-Aside from the message generators, the scheme uses two additional generators: `H_s` and `H_d`. The first (`H_s`), is used for the blinding value (`s`) of the signature. The second generator (`H_d`), is used to sign the signature's domain, which binds both signature and proof to a specific context and cryptographically protects any potential application-specific information (for example, messages that must always be disclosed etc.).
+Aside from the message generators, the scheme uses two additional generators: `H_s` and `H_d`. The first (`H_s`), is used for the blinding value (`s`) of the signature. The second generator (`H_d`), is used to sign the signature's domain, which binds both signature and its derived proofs to a specific context, and cryptographically protects any potential application-specific information (for example, messages that must always be disclosed).
 
-### Encoding of elements to be hashed.
+### Encoding of elements to be hashed
 
 To avoid ambiguity, each element passed to the hash or the xof function, including situations when multiple elements are supplied in a concatenated form, must first be encoded to an appropriate format, depending on its type. Specifically,
 
@@ -290,6 +290,63 @@ As an example of the above transformations, consider the following. Assume that 
 For readability, this document makes these transformations implicitly, but they MUST precede every call to the hash or xof function.
 
 Optional input/parameters to operations that feature in a call to a HASH or XOF function, that are not supplied to the operation should default to an empty octet string. For example, if X is an optional input/parameter that is not supplied, whilst A and B are required, then the procedural step of `HASH(A || X || B)` MUST be evaluated to `HASH(A || "" || B)`.
+
+## Generator point computation
+
+The `compute_generators` function defines how to create a set of generators that form a part of the public parameters used by the BBS Signature scheme to accomplish operations such as Sign, Verify, ProofGen and ProofVerify. It takes one input, the maximum number of selectively revealed messages, which determines the number of generator points to create. The result is a sequence beginning with the `H_s` and `H_d` generators, followed by each of the `H_i` message generators.
+
+This document defines one implementation of this method, `compute_generators_xof`, which uses [@!I-D.irtf-cfrg-hash-to-curve] with an extensible-output hash function to generate a sequence of points with a uniformly random distribution from a seed value and DST.
+
+Each generator creation method must define a unique `GENERATORS_ID` identifier which is incorporated into the ciphersuite ID.
+
+### Security Considerations
+
+TODO
+
+### compute\_generators\_xof
+
+```
+generators = compute_generators_xof(maxMessageCount);
+
+Parameters:
+
+- hash_to_curve_g1: A cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The expand_message method MUST be expand_message_xof, and hash MUST be the same as xof.
+- generator_xof, an extensible-output cryptographic hash function.
+- generator_dst, octet string. The domain separation tag.
+- generator_seed, octet string. A unique seed value for the generator point sequence.
+
+Inputs:
+
+- maxMessageCount (REQUIRED), non-negative integer. The maximum number of selectively-revealed messages to be supported.
+
+Outputs:
+
+- generators, an array of generator points of length `maxMessageCount + 2`.
+
+Procedure:
+
+1. h = generator_xof(generator_seed)
+
+2. length = maxMessageCount + 2
+
+3. for i in 1 to length:
+
+4.    generator_i = Identity_G1
+
+5.    while(generator_i == Identity_G1 or generator_i == P1)
+
+6.        candidate = hash_to_curve_g1(h.read(xof_no_of_bytes), generator_dst)
+
+7.        if candidate not in generators: generator_i = candidate
+
+8. return generators
+```
+
+The `GENERATORS_ID` value for the `compute_generators_xof` generator creation method is assembled using the components of the Suite ID for `hash_to_curve_g1` per the suite naming conventions of [@!I-D.irtf-cfrg-hash-to-curve], excluding the `CURVE_ID` component. The format of the string is:
+
+```
+	HASH_ID || "_" || MAP_ID || "_" || ENC_VAR
+```
 
 ## Operations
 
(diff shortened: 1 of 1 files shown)


BasileiosKal:
Thinking more about it after the call, the only options i can see that would remove the generators from the input of the operations are,
1. Suites will define a `createGenerators` operation (not necessarily the one that is currently on the document) and have each operation call `createGenerators` as part of their procedure, with a note that the output of this function can be cached.
2. Each suite will only define "global" generators but make the "generators" parameter "optional". That way, people can define suites that don't specify the generators and use whatever they want.
3. Define the suite as "dynamic" i.e., give the ability to the ciphersuite to "know" L and/or the issuers PK.

Personally i would go with Option 2. It serves most cases + it keeps the suites "static" and simple. Option 1 will add more steps to each operation, and as I said on the call, I'm not a huge fan of option 3 (i can definitely be convinced otherwise of-course).

 I'm very interested to hear others thoughts on this as well (c.c., @tplooker, @andrewwhitehead ^^).

andrewwhitehead:
I think we need to hammer out the details on generator flexibility. To me it is seeming more like a misfeature that only complicates the API surface and the implementation.

In the original version we had generators that depended on the issuer's public key and the number of messages to be signed, but (to my knowledge) that was only avoiding malleability of those two factors, which is now accomplished by the domain message.

Because this keeps coming up as an issue, and there could conceivably be security considerations or benefits that come along with other creation methods (for example, one that didn't depend on hash-to-curve), I think we ought to leave this open ended in the style of the expand_message function in the hash-to-curve RFC: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#section-5.4.4. Within the RFC there are two expand_message variants defined, and likely every implementation currently uses one or the other. However, each method defines an identifier which is added to the ciphersuite ID, and other arbitrarily-complex methods could be defined in separate RFCs.

BasileiosKal:
Agreed! Just to clarify, you are in favor of Option 1 in [my comment above](https://github.com/decentralized-identity/bbs-signature/pull/161#issuecomment-1141465183) then?? Is that accurate?? 

This would mean that each ciphersuite has a `get_generators` parameter that can be any procedure (like [createGenerators](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-creategenerators)) that deterministically returns a number of distinct, indistinguishable from uniformly random, points in G1. Then each operation will call `get_generators` to create the message generators with a note that the output of this function can be cached??

If I'm not missing something I'm not opposed to that direction. It's the only way i can see that would allow for what the previous version of the spec did (i.e., like `hash_to_curve(PK || i || L)`) without changing the scope of the ciphersuites. Although I'm not 100% convinced that this type of "flexibility" is actually whats needed. This syntax kinda assumes an operation that returns points and discourages (does not disallow) just "picking" random points.

BasileiosKal:
In any case, i think we all agree to remove the generators from the input of the operations, so i suggest moving in in this direction. After that, we can re-define how the generators are picked if need be. 

 @andrewwhitehead i think it's easier to make the update in my [API_update](https://github.com/decentralized-identity/bbs-signature/pull/138) PR?? Is that ok or you prefer making the updates here??

draft-bbs-signatures.md
```diff
@@ -938,22 +966,19 @@ a function that returns the point P in the subgroup G1 corresponding to the cano
...
 
-- blind\_value\_generator\_seed: The seed used to calculate the signature blinding value generator (H_s). Similar to the message\_generator\_seed, there are multiple scopes for the blind\_value\_generator\_seed, with the choices being a global seed, a signer specific seed or a signature specific seed. Also, the ciphersuite MUST define this seed OR how to compute it as a pre-cursor operations to any others.
-
-- sig\_domain\_generator\_seed: The seed for calculating the generator used to sign the signature domain separation tag. The scopes and requirements for this seed are the same as the scopes and requirements of the message\_generator\_seed and blind\_value\_generator\_seed.
+- generator\_xof: The extensible-output cryptographic hash function used in generator computation.
```
BasileiosKal, line 978: I don't see why we would need to use a different hash function for the generators only?? Since `compute_generators_xof` is optional, it seems weird to me to make the ciphersuite to **have** to define the `generator_xof` parameter (or the other `compute_generators_xof`  specific parameters, like `hash_to_curve_g1`).

draft-bbs-signatures.md
```diff
@@ -291,6 +291,63 @@ For readability, this document makes these transformations implicitly, but they
...
+
+5.    while(generator_i == Identity_G1 or generator_i == P1)
+
+6.        candidate = hash_to_curve_g1(h.read(xof_no_of_bytes), generator_dst)
+
+7.        if candidate not in generators: generator_i = candidate
+
+8. return generators
+```
+
```
BasileiosKal, line 344: Since we are making the update, should we maybe consider a different procedure as well?? Something like,
```
Procedure:

1. L= maxMessageCount + 2
2. h_0 = hash(seed || I2OSP(0, 2))
3. h_1 = hash(h_0 || I2OSP(1, 2))
4. for i in 1 to L:
5.    generator_i = Identity_G1
6.    while(generator_i == Identity_G1 or generator_i == P1)
7.        candidate = hash_to_curve_g1(h_i, generator_dst)
8.        h_i = hash(strxor(h_0, h_(i-1)) || I2OSP(i, 2))
9.        if candidate not in generators: generator_i = candidate
10. return generators
```
Here `hash` can be an xof also. What i like about this is that let's say you have cached 1000 generators and you need to sign 1001 messages. In the current procedure you need to compute the 1000 generators from the beginning.  In this approach though, You only need to save the last `h_i` computed, and immediately calculate the next generator. Some notes,

- The input seed is not adversarially controlled so no need for more complicated procedures (because of the guarantees that this function gives, i.e., close to uniform output distribution)
- Is less efficient, but i really don't think that time efficiency is what we should optimize for here. Also, cases like the above will hurt performance. 
- Is usable with whatever hash is defined by the ciphersuite, so no need for extra parameters (this is related with the discussion in #162 as well)

Thoughts?? c.c. @mikelodder7, @tplooker^^

draft-bbs-signatures.md
```diff
@@ -291,6 +291,63 @@ For readability, this document makes these transformations implicitly, but they
...
+Outputs:
+
+- generators, an array of generator points of length `maxMessageCount + 2`.
+
+Procedure:
+
+1. h = generator_xof(generator_seed)
+
+2. length = maxMessageCount + 2
```
BasileiosKal, line 330: Just a though. It may be better to add the +2 when calling the function in the operations?? That way someone that is reading the procedures will not be confused why they got L+2 elements when they asked for L??

draft-bbs-signatures.md
```diff
@@ -291,6 +291,63 @@ For readability, this document makes these transformations implicitly, but they
...
+Parameters:
+
+- hash_to_curve_g1: A cryptographic hash function that takes as an arbitrary octet string input and returns a point in G1 as defined in [@!I-D.irtf-cfrg-hash-to-curve]. The expand_message method MUST be expand_message_xof, and hash MUST be the same as xof.
+- generator_xof, an extensible-output cryptographic hash function.
+- generator_dst, octet string. The domain separation tag.
+- generator_seed, octet string. A unique seed value for the generator point sequence.
+
+Inputs:
+
+- maxMessageCount (REQUIRED), non-negative integer. The maximum number of selectively-revealed messages to be supported.
```
BasileiosKal, line 320:
```suggestion
- maxMessageCount (REQUIRED), non-negative integer. The number of generators to be returned.
```

If the operation is to be called inside each procedure, should this be something like the above?? Otherwise if this input is indeed the maximum number of messages that an issuer can (ever) sign, dosen't this dissallow what the previous version of the spec did (i.e., something like `g_i = hash_to_curve(PK || i || L)`) because in this case it will be weistfull to create maxMessageCount generators??

BasileiosKal (commented):
I really like the direction! Just some thoughts to maybe consider.

draft-bbs-signatures.md
andrewwhitehead: Yes, more likely it would just use the `hash` function. I'm not sure how best to specify the dependency between the hash used in `hash_to_curve_g1` and the other hash used in this method, and also require that it is an XOF. `hash_to_curve_g1`, `generator_seed` etc are meant to be parameters specific to the `compute_generators_xof` method and should probably be in their own section. I'm not sure yet if there is value in supporting `compute_generators` methods that don't necessarily depend on H2C, but maybe using a method that is agnostic of the hash type is good enough for now.

draft-bbs-signatures.md
andrewwhitehead: It's meant to be the maximum number of messages supported by this set of generators, but yes that could be clearer. Having a dependency on L in the seed seems like a liability to me.

BasileiosKal: I agree regarding the seed and L dependency. I would much rather say that the input is a ciphersuite-defined seed. So we don't have to worry a lot about adversarial inputs which can be a pain to deal with (even if we use an XOF only).

IMO there are the following trade-offs in the different input choices to `compute_generators` :

- If we use maxMessageCount, the generators can be a parameter and not created in the procedure of each operation. This makes it clearer that they should be cached.
- On the cons side, it "forces" a hard upper limit to the number of messages that can be signed, without any "technical" reason. It also requires the users to define this number which may not be as "straight forward", since, the more messages they "support", the less efficient the operation will be and the more generators they will need to cache.
- Calling `compute_generators(L)` inside each procedure is clearer IMO and doesn't require more "work" from the implementer to define parameters.
- On the cons side, it may not be clear enough that the output of `compute_generators` should be cached and **NOT** be re-calculated each time.

I would personally go with option 2 and call `compute_generators(L)` inside each operation, with a note that the output must be cached (if it's not, it is not a security issue, just an efficiency one, so i think it's fine).

draft-bbs-signatures.md
BasileiosKal: That's a good point! I can't even imagine what those methods (not h2c based) would look like.

> maybe using a method that is agnostic of the hash type is good enough for now.

Yeap! That's my thinking as well. That's why i took a shot in such a procedure [here](https://github.com/decentralized-identity/bbs-signature/pull/161#discussion_r887355556). Another option can be an hkdf based construction (we already have a "optional dependency" on hkdf for keyGen) or a expand_message based one (since 'compute_generators' depends on h2c either way) as i try to propose [here](https://github.com/decentralized-identity/bbs-signature/pull/161#discussion_r887793786). IMO we best use the same hash for everything. We could use a different hash for `hash_to_curve_g1` by just pointing to the defined h2c suite, but for creating the input seed we should use the bbs+ suite defined `hash`.

Another option can be to make the described `compute_generators` procedure a requirement, and define the `generators_xof` parameter to the ciphersuite as required. However, I'm really not a big fun of this approach.

draft-bbs-signatures.md
andrewwhitehead: I like the idea of using expand_message rather than defining something equivalent, but this method does make it harder to extend a cached list of generators.

BasileiosKal: Good point. Using `expand_message_xmd` you could save `b_0, last_i, b_(last_i)` as state and start from where you left of (and it will be easy to describe in the spec IMO). Using `expand_message_xof` though will be harder. However the current xof based construction has the same issue. So if people want this ability they can always use `expand_message_xmd` with an xof??

draft-bbs-signatures.md
BasileiosKal: +1. In this case i would suggest `maxMessageCount` -> `MessageCount` (or `message_count`). I think this will be clearer.

BasileiosKal: Note, in this case `L` should be inputted to `ProofVerify`. Not really an issue (i know a lot of implementations do this already), just a note so i will not forget to update the inputs.

PR #165, opened 2022-06-04
Editorial updates

BasileiosKal:
Various editorial updates:
1. PK and KeyValidate updates:
    1. Changed `KeyValidate` to accept the de-serialized key (half the times `KeyValidate` was used the PK was (re) de-serialized  immediately after, which is inefficient).
    2. Removed `KeyValidate` from `ProofGen` and `Sign` (we just hash the PK there).
    3. Removed PK de-serialization from `Sign` (the result is not used).
    4. Added PK de-serialization to `ProofVerify`.
 
2. Changed notation from `m~_j1, …, m~_jU` -> `m~_1, …, m~_U` (I think it's clearer this way).

Some updates noted by @dev0x1:
1. Changed pairing equality check in `Verify` to match the one in `ProofVerify`.

2. Fixed typo in `ProofVerify` (`H_s` -> `H_d` in step 8).

I will update/re-base this as required if needed (it will probably have conflicts with some of the other open PRs).


draft-bbs-signatures.md
```diff
@@ -382,27 +380,25 @@ This operation checks if a public key is valid.
 As an optimization, implementations MAY cache the result of KeyValidate in order to avoid unnecessarily repeating validation for known public keys.
 
 ```
-result = KeyValidate(PK)
+result = KeyValidate(W)
 
 Inputs:
 
-- PK (REQUIRED), an octet string of the form outputted by the SkToPk operation.
+- W (REQUIRED), a public key as calculated at step one (1) in SkToPk.
```
tplooker, line 387: I'd prefer we only hand the valid output of SkToPk as the input here

draft-bbs-signatures.md
```diff
@@ -382,27 +380,25 @@ This operation checks if a public key is valid.
...
 Outputs:
 
 - result, either VALID or INVALID.
 
 Procedure:
 
-1. W = octets_to_point_g2(PK)
-
-2. if W is INVALID, return INVALID
+1. if W is INVALID, return INVALID
```
tplooker, line 395: As per above I think we should consider drawing the boundary for this operation as `W` being the valid output from the SkToPk operation

dev0x1:
@BasileiosKal May be we can also update to use consistent ordering of **Abar** and **A'** in proof and input to `hash_to_scalar`, 
in `ProofGen`,
`19. c = hash_to_scalar((PK || Abar || A' || D || C1 || C2 || ph), 1)`

`25. proof = (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU))`

in `ProofVerify`,

`4. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1,...,m^_jU)) = proof`

`10. cv = hash_to_scalar((PK || Abar || A' || D || C1 || C2 || ph), 1)`

draft-bbs-signatures.md
BasileiosKal: I'm not opposed to that. Just to clarify though, do you mean SkToPk not returning octetcs but W directly??

draft-bbs-signatures.md
BasileiosKal: I would prefer that as well. That said it's a trade-off. The problem is that the output of `SkToPk` is an octet string, meaning we will need to de-serialize it in `KeyValidate`. Any time we call `KeyValidate` in a procedure though, we de-serialize the PK again. So we are doing two de-serializations wile we only need one.

Maybe define `W` as a point in `G2` will be fine here??

tplooker: Apologies I meant to say `octets_to_point_g2` instead of `SkToPk` so we would handle that deserialization in the procedure that is going to call `KeyValidate`

tplooker:
> @BasileiosKal May be we can also update to use consistent ordering of **Abar** and **A'** in proof and input to `hash_to_scalar`, in `ProofGen`, `19. c = hash_to_scalar((PK || Abar || A' || D || C1 || C2 || ph), 1)`
> 
> `25. proof = (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1, ..., m^_jU))`
> 
> in `ProofVerify`,
> 
> `4. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_j1,...,m^_jU)) = proof`
> 
> `10. cv = hash_to_scalar((PK || Abar || A' || D || C1 || C2 || ph), 1)`

I would probably raise this as a seperate issue rather than addressing in this PR

tplooker:
@BasileiosKal can you address conflicts?

BasileiosKal:
Updated to remove conflicts + made ph in ProofVerify optional (thanks @dev0x1!!)

BasileiosKal:
After thinking more about it, IMO the best way to de-serialize the PK and check if it is valid is to introduce a `PubkeyToPoint` operation. The reasons are, 

1. If we de-serialize the PK in the procedure and pass the result in KeyValidate directly, we would potentially pass INVALID as input, which I really don’t think it’s wise (not technically wrong, but I’m really afraid of implementation mistakes here).
2. If we de-serialize the PK in KeyValidate, we would need to de-serialize the key twice. 
3. We want the two operations distinct so we can say that people can skip (or cache) the KeyValidate check.

The proposal goes around these issues. Thoughts??

tplooker:
Can you resolve conflicts with this PR?

tplooker (approved):

tplooker:
Discussed on the WG call resolved to make a few changes

- rename pubkey_to_point to octets_to_pubkey
- add identity check to octets_to_pubkey and then remove KeyValidate
- add security consideration for sub group checks for the octet_to_point_g* functions

tplooker (changes requested):
Dropping approval considering the changes we agreed on the WG call

BasileiosKal:
> Discussed on the WG call resolved to make a few changes

Updated and ready for re-review!

draft-bbs-signatures.md
```diff
@@ -503,21 +467,19 @@ Procedure:
...
 
 8. B = P1 + H_s * s + H_d * domain + H_1 * msg_1 + ... + H_L * msg_L
 
-9. C1 = e(A, W + P2 * e)
-
-10. C2 = e(B, P2)
+9. if e(A, W + P2 * e) * e(B, -P2) != 1, return INVALID
 
-11. return C1 == C2
+10. return Valid
```
tplooker, line 482:
```suggestion
10. return VALID
```

draft-bbs-signatures.md
```diff
@@ -993,11 +955,44 @@ Procedure:
...
 
-It is RECOMMENDED for any operation in [BBS Scheme Definition](#bbs-scheme-definition) involving public keys that do not have an explicit invocation to the KeyValidate operation (#keyvalidate) documented in their procedure, that this check be performed prior to executing the operation. An example of where this recommendation applies is the sign (#sign) operation. An example of where an explicit invocation to the KeyValidate operation (#keyvalidate) is already defined and therefore required is the verify (#verify) operation.
+It is RECOMENDED for any operation in [Core Operations](#core-operations) involving public keys, that they deserialize the public key first using the [octets\_to\_pubkey](#octets-to-public-key) operation, even if they only require the octet-string representation of the public key. If the `octets_to_pubkey` procedure returns INVALID the calling operation should also return INVALID and abort. An example of where this recommendation applies is the [Sign](#sign) operation. An example of where an explicit invocation to the `octets_to_pubkey` operation is already defined and therefore required is the [Verify](#verify) operation.
+
+## Valid de-serialization
```
tplooker, line 993:
```suggestion
## Point de-serialization
```

christianpaquin (approved):

PR #174, opened 2022-06-15
Updated section refs to new v16 hash-to-curve spec

christianpaquin:
New v16 of the hash-to-curve spec removed section 5.1, so we need to adapt the section 5.x references in ours. See [v15 and v16 diff](https://www.ietf.org/rfcdiff?difftype=--hwdiff&url2=draft-irtf-cfrg-hash-to-curve-16.txt).

I couldn't find the actual reference to the spec in the Normative References section, only generic `[@!I-D.irtf-cfrg-hash-to-curve]` refs; I assume the build system turns that to a ref to the latest version?


tplooker (approved):
LGTM thanks @christianpaquin and yes the reference resolution I believe auto resolves to the latest draft available for that name

tplooker:
Minor editorial tweak, merging

PR #175, opened 2022-06-15
Use Identity_G1 instead of 1 in ProofVerify

christianpaquin:
Use Identity_G1 instead of 1 in ProofVerify

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -671,9 +671,9 @@ Procedure:
 
 13. if c != cv, return INVALID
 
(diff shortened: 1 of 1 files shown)


tplooker:
> Identity_GT I suppose?

Yeah this would make more sense to me? could we add that to the terminology section @christianpaquin

tplooker (approved):

andrewwhitehead (approved):

tplooker:
Multiple approvals, minor editorial change that provides an important clarification to the ProofVerify procedure, merging.

PR #177, opened 2022-06-16
Update generator creation procedure

andrewwhitehead:
This removes several of the ciphersuite parameters in favor of a generator creation procedure which reuses the `expand_message` operation defined by hash-to-curve suite. Additional PRs would be needed to update the use of generators in the suite's operations, and to update the test vectors. I'm trying to avoid conflicts with open PRs.

The operation currently defines a `hash_output_bytes` value, but this would likely switch to `expand_length` used in #151.

Supersedes #161 and the discussion on that PR.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -729,7 +729,7 @@ Procedure:
 
 9.        if candidate not in generators[1..i-1]: generator_i = candidate
 
-10. return generators
+10. return (generator_1, ..., generator_count)
 ```
 
 ## MapMessageToScalar
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -711,9 +711,9 @@ Outputs:
 
 Procedure:
 
-1. n = 1
+1. v = expand_message(generator_seed, seed_dst, hash_output_bytes)
 
-2. v = expand_message(generator_seed, seed_dst, hash_output_bytes)
+2. n = 1
 
 3. for i in range(1, count):
 
@@ -721,13 +721,13 @@ Procedure:
 
 5.    while generator_i == Identity_G1 or generator_i == P1:
 
-6.        candidate = hash_to_curve_g1(v, generator_dst)
+6.        v = expand_message(v || I2OSP(n, 4), seed_dst, hash_output_bytes)
 
-7.        if candidate not in generators[1..i-1]: generator_i = candidate
+7.        n = n + 1
 
-8.        v = expand_message(v || I2OSP(n, 4), seed_dst, hash_output_bytes)
+8.        candidate = hash_to_curve_g1(v, generator_dst)
 
-9.        n = n + 1
+9.        if candidate not in generators[1..i-1]: generator_i = candidate
 
 10. return generators
 ```
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -249,7 +249,7 @@ In definition of this signature scheme there are two possible variations based u
 
 Throughout the operations of this signature scheme, each message that is signed is paired with a specific generator (point in G1). Specifically, if a generator `H_1` is multiplied with `msg_1` during signing, then `H_1` MUST be multiplied with `msg_1` in all other operations as well (signature verification, proof generation and proof verification). For simplicity, each function will take as input the list of generators to be used with the messages. Those generators can be any distinct element from the generators list `H`. Applications for efficiency can elect to pass the indexes of those generators to the list `H` instead. Care must be taken for the correct generator to be raised to the correct message in that case.
 
-Aside from the message generators, the scheme uses two additional generators: `H_s` and `H_d`. The first (`H_s`), is used for the blinding value (`s`) of the signature. The second generator (`H_d`), is used to sign the signature's domain, which binds both signature and proof to a specific context and cryptographically protects any potential application-specific information (for example, messages that must always be disclosed etc.).
+Aside from the message generators, the scheme uses two additional generators: `H_s` and `H_d`. The first (`H_s`), is used for the blinding value (`s`) of the signature. The second generator (`H_d`), is used to sign the signature's domain, which binds both the signature and its derived proofs to a specific context and cryptographically protects any potential application-specific information (for example, messages that must always be disclosed etc.).
 
 ## Encoding of elements to be hashed
 
@@ -680,18 +680,30 @@ Procedure:
 
 # Utility Operations
 
-## CreateGenerators
+## Generator point computation
 
-The CreateGenerators operation defines how to create a set of generators that form a part of the public parameters used by the BBS Signature scheme to accomplish operations such as Sign, Verify, ProofGen and ProofVerify.
+The `create_generators` operation defines how to create a set of generators that form a part of the public parameters used by the BBS Signature scheme to accomplish operations such as Sign, Verify, ProofGen and ProofVerify. It takes one input, the number of generator points to create, which is determined in part by the number of signed messages.
+
+As an optimization, implementations MAY cache the result of `create_generators` for a specific `generator_seed` (determined by the ciphersuite) and `count`. The values `n` and `v` MAY also be cached in order to efficiently extend a existing list of generator points.
 
 ```
-generators = CreateGenerators(dst, message_generator_seed, length);
+generators = create_generators(count)
 
 Inputs:
 
-- dst (REQUIRED), octet string. Domain Separation Tag.
-- message_generator_seed (REQUIRED), octet string.
-- length (REQUIRED), unsigned integer. Number of generators to create from the seed and dst.
+- count (REQUIRED), unsigned integer. Number of generator points to create.
+
+Parameters:
+
+- generator_seed, octet string. A seed value selected by the ciphersuite.
+- hash_to_curve_g1. A hash-to-curve suite selected by the ciphersuite.
+
+Definitions:
+
+- expand_message is the corresponding operation associated with the ciphersuite's hash_to_curve_g1 parameter.
+- seed_dst is the octet string representing "BBS-SIG-GENERATOR-SEED" in the ASCII character encoding.
+- hash_output_bytes is the integer given by ceil((log2(r) + k) / 8).
+- generator_dst is the octet string representing "BBS-SIG-GENERATOR-DST" in the ASCII character encoding.
 
 Outputs:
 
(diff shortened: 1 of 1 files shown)


draft-bbs-signatures.md
```diff
@@ -680,38 +680,56 @@ Procedure:
...
 Outputs:
 
 - generators, an array of generators.
 
 Procedure:
 
-1. h = xof(seed)
+1. n = 1
+
+2. v = expand_message(generator_seed, seed_dst, hash_output_bytes)
```
andrewwhitehead, line 716: For an XOF hash, this is not much different from the previous procedure, it only adds a DST.

draft-bbs-signatures.md
```diff
@@ -680,38 +680,56 @@ Procedure:
...
+
+- generator_seed, octet string. A seed value selected by the ciphersuite.
+- hash_to_curve_g1. A hash-to-curve suite selected by the ciphersuite.
+
+Definitions:
+
+- expand_message is the corresponding operation associated with the ciphersuite's hash_to_curve_g1 parameter.
+- seed_dst is the octet string representing "BBS-SIG-GENERATOR-SEED" in the ASCII character encoding.
+- hash_output_bytes is the integer given by ceil((log2(r) + k) / 8).
+- generator_dst is the octet string representing "BBS-SIG-GENERATOR-DST" in the ASCII character encoding.
```
andrewwhitehead, line 706: I don't see any need for the generator DST and seed DST to be ciphersuite parameters, although the actual values are up for debate (these are chosen to be similar to the KeyGen salt).

draft-bbs-signatures.md
alessandroguggino, line 716: Why not use an `xof` instead of `expand_message` and then call `.read(...)` inside the for cycle?

andrewwhitehead: This procedure is meant to be easily implemented whether the hash-to-curve suite's hash function is extensible-output or not.

draft-bbs-signatures.md
```diff
@@ -680,38 +680,56 @@ Procedure:
...
+Parameters:
+
+- generator_seed, octet string. A seed value selected by the ciphersuite.
+- hash_to_curve_g1. A hash-to-curve suite selected by the ciphersuite.
+
+Definitions:
+
+- expand_message is the corresponding operation associated with the ciphersuite's hash_to_curve_g1 parameter.
+- seed_dst is the octet string representing "BBS-SIG-GENERATOR-SEED" in the ASCII character encoding.
+- hash_output_bytes is the integer given by ceil((log2(r) + k) / 8).
```
BasileiosKal, line 705: We don't mod something here so i don't see the reason for this suggestion. This will be the seed that will go into hash_to_curve which can be whatever length we want wright?? Or are we going to use that value, just so we will not introduce an extra variable?? It still seems a little extensive though, why not use `ceil(log2(r)/8)` for example??

draft-bbs-signatures.md
```diff
@@ -680,38 +680,56 @@ Procedure:
 
 # Utility Operations
 
-## CreateGenerators
+## Generator point computation
 
-The CreateGenerators operation defines how to create a set of generators that form a part of the public parameters used by the BBS Signature scheme to accomplish operations such as Sign, Verify, ProofGen and ProofVerify.
+The `create_generators` operation defines how to create a set of generators that form a part of the public parameters used by the BBS Signature scheme to accomplish operations such as Sign, Verify, ProofGen and ProofVerify. It takes one input, the number of generator points to create, which is determined in part by the number of selectively-revealed signed messages.
+
+As an optimization, implementations MAY cache the result of `create_generators` for a specific `generator_seed` (determined by the ciphersuite) and `count`. The values `c` and `v` MAY also be cached in order to efficiently extend a existing list of generator points.
```
BasileiosKal, line 687: Is `c` here `count`??

draft-bbs-signatures.md
BasileiosKal: That's true but since this procedure will be used to create ciphersuite specific operators i like not adding them to the things an implimenter will need to choose. Maybe adding them as operation definitions??

draft-bbs-signatures.md
BasileiosKal: There is a difference IMO and that is that we re-initialize the xof every time with a different seed. Personally, I’m fine with the approach but an alternative could be to call expand_message only once to get all the bytes necessary and only re-call it when we get a duplicate or a point at infinity.

draft-bbs-signatures.md
BasileiosKal: And actually, thinking more about it, does this method "require" expand_message?? Wouldn't just `hash` do, where if `hash` is an xof, we will specify the output length??

draft-bbs-signatures.md
andrewwhitehead: Yes I wasn't clear there, it's only similar for the first generator.

draft-bbs-signatures.md
andrewwhitehead: Oops, I renamed that to `n`.

draft-bbs-signatures.md
andrewwhitehead: > Maybe adding them as operation definitions??

They are added as definitions currently

draft-bbs-signatures.md
andrewwhitehead: `hash_output_bytes` would likely be replaced by `expand_length` in future, but that's in another PR. I didn't want to reference `xof_no_bytes` here.

andrewwhitehead: > And actually, thinking more about it, does this method "require" expand_message?? Wouldn't just hash do, where if hash is an xof, we will specify the output length??

It could but that would limit `v` to the hash's output size for fixed output hashes. I prefer using `expand_message` here because it's a primitive that is guaranteed to be available and should be hard to misuse.

draft-bbs-signatures.md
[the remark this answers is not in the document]
andrewwhitehead: That could be a little harder to implement, because generator_seed doesn't have a bounded length, so the implementer might need to allocate a separate buffer to create the input for expand_message.

draft-bbs-signatures.md
```diff
@@ -680,38 +680,56 @@ Procedure:
...
+5.    while generator_i == Identity_G1 or generator_i == P1:
 
-4.    while(generator_i == Identity_G1 or generator_i == P1)
+6.        candidate = hash_to_curve_g1(v, generator_dst)
 
-5.        candidate = hash_to_curve_g1(h.read(xof_no_of_bytes), dst)
+7.        if candidate not in generators[1..i-1]: generator_i = candidate
 
-6.        if candidate not in generators: generator_i = candidate
+8.        v = expand_message(v || I2OSP(n, 4), seed_dst, hash_output_bytes)
```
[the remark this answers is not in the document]
andrewwhitehead: > generator_2 is not indistinguishable from random if you know generator_1

Well, this assumes that generator_1 is uniquely determined from the generator_seed, but `v` could be made large enough to ensure there are collisions in the hash-to-curve output. The generator points are not used as inputs to create the next state.

Another option would be to use `v[X..]` as the hash-to-curve input, so the first X bytes remain as 'private' for the generator state machine.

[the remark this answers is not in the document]
andrewwhitehead: Here's what the last option might look like:

```
private_bytes = 16  # could be in definitions, probably does not need to be curve dependent
hash_output_bytes = private_bytes + expand_length
n = 1
v = expand_message(generator_seed, seed_dst, hash_output_bytes)
for i in range(1, count):
   generator_i = Identity_G1
   while generator_i == Identity_G1 or generator_i == P1:
       candidate = hash_to_curve_g1(v[private_bytes..hash_output_bytes], generator_dst)
       if candidate not in generators[1..i-1]: generator_i = candidate
       v = expand_message(v || I2OSP(n, 4), seed_dst, hash_output_bytes)
       n = n + 1
return generators
```

Another option is to initialize v[..private_bytes] with zeroes when hashing the seed.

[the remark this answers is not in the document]
BasileiosKal: > generator_i does not determine the generator_(i+1) though, as the generator points are not used as inputs to create the next state. There are multiple v values which could produce each generator point, especially if we increase the size of v. And of course you would have to invert the hash somehow to predict the next value.

Yea this is true in practice, however it may not be "strong" enough for a security proof. Proving "statistical closeness to uniform" is hard, since`P(generator_i | generator_(i-1), ..., generator_1) >> 1/i`. Also, I think we should treat the seed as a known, fixed value (like an IV) in this case, which does not help with the proof. Even if that was not the case though, it is true that multiple `v`'s can give the same point, but a simulator cannot find them, if we assume coalition resistance. As you noted though, we may be able to prove this for "computationally close to uniform".

Personally, I like the cut-off method. A prefix-free method could also work and is pretty close to the current direction of the PR. That said, I do suggest leaving any updates (if there are any required) after we do a formal analysis. I will try to write down a formal proof, and if there are any updates required as a result, we can make them in following PRs.

draft-bbs-signatures.md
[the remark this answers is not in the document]
BasileiosKal: Good point. Maybe then calculating a first `v` as `v = expand_message(generator_seed, seed_dst, hash_output_bytes)` and then again as `v = expand_message(v || I2OSP(0, 4), seed_dst, hash_output_bytes)` (this second `v` will be the "seed" for the first generator)?? I do realize this adds complexity, but it will make it easier to move to a cut-off or prefix-free like construction if we decide to do so.

draft-bbs-signatures.md
BasileiosKal: Ok. Thats a good point and i think that you are right in that expand_message will be just as available in practice.

draft-bbs-signatures.md
BasileiosKal: Oh, you right. My bad!

draft-bbs-signatures.md
```diff
@@ -680,38 +680,56 @@ Procedure:
...
-5.        candidate = hash_to_curve_g1(h.read(xof_no_of_bytes), dst)
+7.        n = n + 1
 
-6.        if candidate not in generators: generator_i = candidate
+8.        candidate = hash_to_curve_g1(v, generator_dst)
 
-7. return generators
+9.        if candidate not in generators[1..i-1]: generator_i = candidate
+
+10. return generators
```
BasileiosKal, line 732:
```suggestion
9.        if candidate not in (generator_1, ..., generator_(i-1)):

10.              generator_i = candidate

11. return (generator_1, ..., generator_n)
```
Just a suggestion because we "technically" don’t define the generators list

BasileiosKal (approved):
Just a minor comment. LGTM otherwise!!

draft-bbs-signatures.md
andrewwhitehead: Thanks, maybe it should use `n` instead of `count` as the length parameter in line with hash-to-scalar?

BasileiosKal: Oh right, sorry! Yea I’m fine with both honestly. I'm updating hash-to-scalar either way, so I can make that consistent with what we do here. I updated the suggestion to use `count` for now, just for ease, but I can go with either.

tplooker (approved):

tplooker:
Discussed on WG call, multiple reviews and approvals, merging, thanks @andrewwhitehead!

Issue #178, opened 2022-06-20
Align language consistency disclose vs reveal


tplooker:
We use both the term disclosed messages and revealed messages throughout the draft, considering we offer in the introduction that selective disclosure is one of the key properties enabled by the scheme I suggest we change all usages of the term reveal to disclose in the context of messages.

tplooker:
Somewhat blocked by a resolution to #138 as this makes some modifications to the API.

tplooker:
Closed with merging of #183

Issue #168, opened 2022-06-10
Explicitly give the number of bytes for the PRF function


christianpaquin:
The PRF function in `ProofGen` takes the value `8*ceil(log2(r))`. It would be helpful to explicitly define its value for the BLS12-381 ciphersuite to avoid guess work. I assume the value is 256.

tplooker:
Suggestion discussed on WG call is to generalize the `xof_no_of_bytes` parameter for this purpose and parameterize it in the proof gen procedure.

Issue #159, opened 2022-05-27
API Update


BasileiosKal:
Trying to update the API has resulted in mainly two proposals from @andrewwhitehead and @tplooker. The first one using boolean variables to distinquish between revealed and un-revealed messages and the second one using indexes. You can compare the 2 APIs [here](https://github.com/BasileiosKal/dif-bbs-signature-api):
- [API_1](https://github.com/BasileiosKal/dif-bbs-signature-api/tree/main/API_1_Boolean) (using Booleans): [Sign](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_1_Boolean/draft-bbs-signatures.md#sign), [Verify](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_1_Boolean/draft-bbs-signatures.md#verify), [ProofGen](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_1_Boolean/draft-bbs-signatures.md#proofgen), [ProofVerify](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_1_Boolean/draft-bbs-signatures.md#proofverify)
- [API_2 ](https://github.com/BasileiosKal/dif-bbs-signature-api/tree/main/API_2_Indexes)(using Indexes): [Sign](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_2_Indexes/draft-bbs-signatures.md#sign), [Verify](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_2_Indexes/draft-bbs-signatures.md#verify), [ProofGen](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_2_Indexes/draft-bbs-signatures.md#proofgen), [ProofVerify](https://github.com/BasileiosKal/dif-bbs-signature-api/blob/main/API_2_Indexes/draft-bbs-signatures.md#proofverify)

We can somewhat compare the two APIs based on the total number of lines (i.e., everything inside each code block) and the total number of procedure steps,

### API 1 (Booleans)

|                  |  lines   | Procedure Steps | 
| ----------- | ------ | ----------------- | 
| Sign           |   54    |        9                   | 
| Verify         |  59     |       11                  | 
| ProofGen   |  109   |       26                  | 
| ProofVerify |   78    |       14                  | 

### API 2 (Indexes)

|                   | lines | Procedure Steps |
| ----------   | ---- | ----------------- |
[...]

PR #183, opened 2022-06-22
New API update

### LINKED ISSUE #163: Structured vs de-structured inputs
Lately the direction the document is going, is to pass inputs in a structured form (i.e., `octets_to_signature(signature)` vs `octets_to_signature(A, e, s)` etc.). Related to #138 and #155.

Opening the issue so we can keep track of the update and make notation consistent across the spec.

BasileiosKal:
Substitutes previous PR #138 on API update since it had more conflicts than text at this point.

### Updates:
This PR introduces the following main changes:

1. Removes the generators from the input of the operations. Now the generators are computed as part of the operation procedure, using `create_generators` and a ciphersuite defined seed.
2. Makes all arguments “structured” (this is to be more explicit when for example we need a specific subarray of an inputted array in the correct order, i.e., revealed messages from messages using the indexes etc.). This closes #163.
3. Introduces an additional section for the core operations called `Precomputations`. The reason is that, as the result of the above updates, a lot of steps are needed to de-structure the inputs or get sub-arrays of the inputted arrays etc. I find that those steps “fill” the main procedure, making it harder to read (especially ProofGen/ ProofVerify). So, I moved them to the “Precomputation” section to try and make things more readable. 

### Pros and Cons:
**Pros**:
-	We are as explicit as it gets when it comes to the ordering of things etc.
-	The operations api is much nicer IMO (no more generators there!! Thanks @andrewwhitehead!!)

**Cons**:
-	A lot more steps. Been explicit requires describing every step of how we get certain elements.

### Note:
Regarding the `Precomputations` section if the WG finds it to be more confusing than helpful I’m fine with removing it. Personally, I like it, but it was an “experiment” to see how it looks.

Also addresses #159, #117, #113, #168, #139


BasileiosKal:
Also made the `revealed` -> `disclosed` update here. Closes #178.

alessandroguggino (commented):
Great!

I would specify why `R` and `U` are called that.
Maybe something like:

```
- R, is the non-negative integer representing the number of disclosed
     (revealed) messages, i.e., R = length(disclosedIndexes).
     If no messages are disclosed, the value of R MUST evaluate to zero (0).
- U, is the non-negative integer representing the number of hidden
     (undisclosed) messages, i.e., U = L - R.
```

Also, why not add in the precomputations the `MapMessageToScalarAsHash` operation?

draft-bbs-signatures.md
```diff
@@ -389,78 +387,102 @@ Procedure:
...
-- msg_1,..., msg_L (OPTIONAL), a vector of octet strings.
-- H_1,..., H_L (OPTIONAL), points of G1. Generators used to commit each message.
+- SK (REQUIRED), a non negative integer mod r outputted by the KeyGen
+                 operation.
+- PK (REQUIRED), an octet string of the form outputted by the SkToPk
+                 operation provided the above SK as input.
+- header (OPTIONAL), an octet string containing context and application
+                     specific information. If not supplied, it defaults
+                     to an empty string.
+- messages (OPTIONAL), a vector of octet strings.
```
OR13, line 401: is this really optional?

draft-bbs-signatures.md
```diff
@@ -389,78 +387,102 @@ Procedure:
...
-7. return signature_octets
+8. return signature_octets
 ```
 
 **Note** When computing step 7 of the above procedure there is an extremely small probability (around `2^(-r)`) that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evalute to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 7, and or `A != Identity_G1` after step 7 to prevent the production of invalid signatures.
 
 ### Verify
 
-This operation checks that a signature is valid for a given header and vector of messages against a supplied public key (PK).
+ This operation checks that a signature is valid for a given header and vector of messages against a supplied public key (PK). The messages MUST be supplied in this operation in the same order they were supplied to [Sign](#sign) when creating the signature.
```
OR13, line 448: excellent

draft-bbs-signatures.md
```diff
@@ -473,52 +495,80 @@ Procedure:
...
+- signature (REQUIRED), an octet string of the form outputted by the
+                        Sign operation.
+- header (OPTIONAL), an octet string containing context and application
+                     specific information. If not supplied, it defaults
+                     to an empty string.
 - ph (OPTIONAL), octet string containing the presentation header.
-- msg_1,..., msg_L (OPTIONAL), octet strings. Messages in input to Sign.
-- H_1,..., H_L (OPTIONAL), points of G1. The generators in input to Sign.
-- RevealedIndexes (OPTIONAL), vector of unsigned integers. Indexes of revealed messages.
+- disclosedIndexes (OPTIONAL), vector of unsigned integers in ascending
```
OR13, line 530: if order matters should this be optional?

OR13 (approved):

draft-bbs-signatures.md
BasileiosKal: "Technically" yes. It is optional because someone could only sign the header (with always disclosed information) and not the messages (which you can selectively disclose from) and vice versa.

If they sign neither the result will be a signature on just the signers PK, which "technically" is valid (but most likely useless). To support the other 2 use cases though we left both as optional.

draft-bbs-signatures.md
BasileiosKal: Thank you!! 😄

draft-bbs-signatures.md
BasileiosKal: Thank you very much for your feedback!! This can be confusing, but I think having it as REQUIRED will be like "forcing" the prover to disclose messages. So, I don't see a better alternative using the current syntax. I’ve added an `if not supplied it defaults to the empty array "()"` to try and make it clearer. 

In general, what I want to enforce is that if not supplied, no revealed messages can be supplied either. However, if it is supplied, it must be ordered.

andrewwhitehead: It seems to me like messages should come before disclosedIndexes in the argument list if only because of the optionality... if there are no signed messages then there are definitely no disclosedIndexes, but the same can't be said for the reverse.

draft-bbs-signatures.md
```diff
@@ -527,119 +583,153 @@ Procedure:
 
 3. (A, e, s) = signature_result
 
-4. (i1, i2,..., iR) = RevealedIndexes
+4. gen_octets =  (H_s || H_d || H_1 || ... || H_L)
 
-5. (j1, j2,..., jU) = [L] \ RevealedIndexes
+5. domain_prime = (PK || L || gen_octets || Ciphersuite_ID || header)
```
andrewwhitehead, line 588: It's less clear now that the hash input encoding rules apply here. I think we need to define a helper function like `encode_for_hash` and explain the encoding rules there. That could be a separate update, though.

draft-bbs-signatures.md
```diff
@@ -527,119 +583,153 @@ Procedure:
 
 3. (A, e, s) = signature_result
 
-4. (i1, i2,..., iR) = RevealedIndexes
+4. gen_octets =  (H_s || H_d || H_1 || ... || H_L)
```
andrewwhitehead, line 586:
```suggestion
4. gen_octets = (H_s || H_d || H_1 || ... || H_L)
```

draft-bbs-signatures.md
```diff
@@ -527,119 +583,153 @@ Procedure:
...
+5. domain_prime = (PK || L || gen_octets || Ciphersuite_ID || header)
 
-6. generators =  (H_s || H_d || H_1 || ... || H_L)
+6. domain = hash_to_scalar(domain_prime, 1)
 
-7. domain = hash_to_scalar((PK || L || generators || Ciphersuite_ID || header), 1)
+7. (r1, r2, e~, r2~, r3~, s~) = hash_to_scalar(PRF(prf_len), 6)
 
-8. (r1, r2, e~, r2~, r3~, s~) = hash_to_scalar(PRF(8*ceil(log2(r))), 6)
+8. (m~_j1, ..., m~_jU) =  hash_to_scalar(PRF(prf_len), U)
```
andrewwhitehead, line 594:
```suggestion
8. (m~_j1, ..., m~_jU) = hash_to_scalar(PRF(prf_len), U)
```

draft-bbs-signatures.md
```diff
@@ -527,119 +583,153 @@ Procedure:
...
+
+2. (j1, ..., jU) = range(1, L) \ disclosedIndexes
+
+3. msg_i1, ..., msg_iR = disclosedMessages[1], ..., disclosedMessages[R]
+
+4. (H_s, H_d, MsgGenerators) = create_generators(generator_seed, L+2)
+
+5. H_1, ..., H_L = MsgGenerators[1], ..., MsgGenerators[L]
+
+6. H_i1, ..., H_jR = MsgGenerators[i1], ..., MsgGenerators[iR]
```
andrewwhitehead, line 690:
```suggestion
6. H_i1, ..., H_iR = MsgGenerators[i1], ..., MsgGenerators[iR]
```

draft-bbs-signatures.md
tplooker: Yeah I agree with this ^

draft-bbs-signatures.md
```diff
@@ -473,52 +497,84 @@ Procedure:
 
 5. if W is INVALID, return INVALID
 
-6. generators =  (H_s || H_d || H_1 || ... || H_L)
+6. gen_octets =  (H_s || H_d || H_1 || ... || H_L)
```
andrewwhitehead, line 500:
```suggestion
6. gen_octets = (H_s || H_d || H_1 || ... || H_L)
```

draft-bbs-signatures.md
```diff
@@ -389,78 +387,104 @@ Procedure:
...
-- L, is the non-negative integer representing the number of messages to be signed e.g length(msg_1,...,msg_L).  Note if no messages are supplied as an input to this operation, the value of L MUST evaluate to zero (0).
+- L, is the non-negative integer representing the number of messages to
+     be signed e.g length(messages). If no messages are supplied as an
+     input, the value of L MUST evaluate to zero (0).
 
 Outputs:
 
 - signature, a signature encoded as an octet string.
 
+Precomputations:
```
tplooker, line 420: Just a note I think we should provide an explanation some where as to why we are adding another concept to the definition of an operation. I assume this is because the result of such could be cached or handled in an alternative manner?

draft-bbs-signatures.md
BasileiosKal: That's true. I would prefer taking this up in another PR as well though. Opened #185 to keep track of this. Thanks for raising.

draft-bbs-signatures.md
BasileiosKal: Done!

draft-bbs-signatures.md
BasileiosKal: The real reason is for readability. In practice it's a "setting up" process where we can get all necessary elements for the "core procedure". So essentially yes, staff that can be cached and that "cannot" return INVALID.

Added a small paragraph in the start of the Core operations section. We can always update that as well.

andrewwhitehead (approved):

tplooker (approved):

tplooker:
Multiple reviews and approvals, merging, thanks @BasileiosKal!

PR #184, opened 2022-06-23
Clarified scalar operations

christianpaquin:
Added explicit mod r and moved the scalar multiplications before additions to avoid op-priority confusions.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -549,7 +549,7 @@ Procedure:
 
 14. D = B * r1 + H_s * r2
 
-15. s' = s + r2 * r3
+15. s' = r2 * r3 + s mod r
 
 16. C1 = A' * e~ + H_s * r2~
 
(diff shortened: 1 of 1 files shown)


andrewwhitehead (approved):

BasileiosKal (approved):

BasileiosKal:
LGTM. 

Just a note. maybe we are encouraging people to the x*y + z operations and then mod r?? which I don't think is a good idea in practice (it should be ((x * y mod r) + z) mod r ??). 

Another option would be to define that all operations between scalars happen in the multiplicative group mod r?? That would be the more "mathematically correct" description, although I’m uncertain if it would be the clearer one. 

I'm definitely overthinking this though. I think we should move with merging this 😄

christianpaquin:
> Just a note. maybe we are encouraging people to do the x*y + z operations and then mod r?? which I don't think is a good idea in practice (it should be ((x * y mod r) + z) mod r ??).

It's true, but it also becomes cumbersome. Your suggestion for a note somewhere that all calculations in that field is mod r. I simply wanted to avoid someone leaving a larger scalar and then trying to multiply that to the elliptic curve point; result would be validate but would be very inefficient. A simple mod r at the end indicates it's ok to apply the modulo at the end of the equation (confirming my interpretation of the spec).

tplooker (approved):

tplooker:
Multiple reviews and approvals, merging.

PR #186, opened 2022-06-24
Update KeyGen procedure to use hash_to_scalar

andrewwhitehead:
Since we don't use HKDF anywhere else and aren't trying to be compatible with the BLS signatures spec (different salt).

It's a bit odd using camel case for the input parameters and snake case for the parameter, but I'm trying to be consistent with the rest of the spec.

Likewise, I'm not sure if we prefer len() or length(), and either way we might need to define that.


BasileiosKal:
Although i agree with removing the dependency to hkdf I'm worried that we are changing the security assumptions.

It's true that the H2C spec claims `expand_message` to be intistinquishable from a RO, however 1. there is no formal published work proving that (the claim is based in comparisons with [[CDMP05]](https://link.springer.com/content/pdf/10.1007/11535218_26.pdf)), 2. we didn't required that up until now. We only required `expand_message` to have basic "hash" properties i.e., "un-revershability" and "coalition resistance". So using it for `KeyGen` does raise the amount of assumptions we have to make.

### Note:
This is probably completely fine. We can always add this to the list of things we need to revisit and do a formal analysis on ;P. So I'm happy to move with this.

draft-bbs-signatures.md
```diff
@@ -304,58 +304,45 @@ The schemes operations defined in this section depend on the following parameter
...
 For security, IKM MUST be infeasible to guess, e.g. generated by a trusted source of randomness.
 
-IKM MUST be at least 32 bytes long, but it MAY be longer.
+keyMaterial MUST be at least 32 bytes long, but it MAY be longer.
 
-Because KeyGen is deterministic, implementations MAY choose either to store the resulting SK or to store IKM and call KeyGen to derive SK when necessary.
+Because KeyGen is deterministic, implementations MAY choose either to store the resulting SK or to store keyMaterial and call KeyGen to derive SK when necessary.
 
-KeyGen takes an optional parameter, key\_info. This parameter MAY be used to derive multiple independent keys from the same IKM.  By default, key\_info is the empty string.
+KeyGen takes an optional input, keyInfo. This parameter MAY be used to derive multiple independent keys from the same key material.
```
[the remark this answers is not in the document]
andrewwhitehead: There is a similar comment after the method but it could be updated.

andrewwhitehead:
For reference, the VOPRF draft also uses hash-to-scalar for key generation: https://cfrg.github.io/draft-irtf-cfrg-voprf/draft-irtf-cfrg-voprf.html#name-key-generation-and-context-

christianpaquin (approved):

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -285,53 +285,48 @@ Those rules will be used explicitly on every operation. See also [Serialize](#se
...
-5.     PRK = HKDF-Extract(salt, IKM || I2OSP(0, 1))
-6.     OKM = HKDF-Expand(PRK, key_info || I2OSP(L, 2), L)
-7.     SK = OS2IP(OKM) mod r
-8. return SK
+1.  if length(key_material) < 32, return INVALID
+2.  if length(key_info) > 65535, return INVALID
+3.  derive_input = key_material || I2OSP(length(key_info), 2) || key_info
+4.  counter = 0
+5.  SK = 0
+6.  while SK == 0:
```
BasileiosKal, line 323: Just a note: `hash_to_scalar` currently guarantees that `SK !- 0`. IMO of course it would be better to update `hash_to_scalar` than this operation.

Wind4Greg (approved):

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -285,53 +285,48 @@ Those rules will be used explicitly on every operation. See also [Serialize](#se
...
+7.      if counter > 255, return INVALID
+8.      SK = hash_to_scalar(derive_input || I2OSP(counter, 1), 1, key_dst)
+9.      counter = counter + 1
+10. return SK
```
BasileiosKal, line 327:
```suggestion
8.      SK = hash_to_scalar(derive_input || I2OSP(counter, 1), 1, key_dst)
9.      if SK is INVALID, return INVALID
10.     counter = counter + 1
11. return SK
```

After merging #240 `hash_to_scalar` is updated and it may return INVALID. We should check here for it.

BasileiosKal:
Discussed on WG of 14th of Feb. Will discuss after PR #243 is resolved.

BasileiosKal:
Discussed on WG call on 6th of Mar. Will update the test vectors and try to incorporate the update for draft 02.

BasileiosKal:
Discussed on WG call 13th of Mar, will update to include the ciphersuite id to the default dst. Will review after.

tplooker:
Discussed briefly on the 10th April call, @andrewwhitehead believes all pending updates to the schemes definition have been done and we will discuss again next week as a wider group about approval and merging.

Wind4Greg (approved):

BasileiosKal (approved):

tplooker (approved):

tplooker:
Discussed on WG call 19th of June, multiple approvals, merging.

PR #188, opened 2022-06-26
Add the revealed messages to the challenge

### LINKED ISSUE #74: Adding the revealed messages to the chalenge to avoid forgery

BasileiosKal:
Closes #74

Replaces #128. See the discussion there for more context.

Adds the revealed indexes, the number of revealed messages and the revealed messages themselves to the Fiat–Shamir challenge hash.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -427,9 +427,9 @@ Precomputations:
 
 Procedure:
 
-1. gen_octets =  (H_s || H_d || H_1 || ... || H_L)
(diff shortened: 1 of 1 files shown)


tplooker (approved):

andrewwhitehead (approved):

BasileiosKal:
Multiple approvals. Discussed in the WG call. Merging

PR #190, opened 2022-06-26
encode for hash operation

### LINKED ISSUE #74: Adding the revealed messages to the chalenge to avoid forgery

### LINKED ISSUE #185: Elements to be hashed update
The current [Encoding of Elements to be Hashed](https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-encoding-of-elements-to-be-) section needs to be updated. The reason is that elements before they are hashed, they are concatenated with other elements in multiple steps. As @andrewwhitehead noted This is not clearly captured by the section yet.

Some options: 
1. Define a helper `encode_for_hash` function (see @andrewwhitehead's comment [here](https://github.com/decentralized-identity/bbs-signature/pull/183#discussion_r905553694))
2. Say something like `Each time elements are concatenated together, they must first be turned to octet strings. Every time an element other than a scalar, a point or a non-negative integer is transformed to an octet string, the length of the resulting octet string must be appended to itself.`

BasileiosKal:
Closes #185 using @andrewwhitehead's proposal (introducing an encode_for_hash operation).

Opened as a draft because of the timetable + the fact that it is not a small pr. Will mark it as ready to review if the other open PRs are merged.

### NOTES:
1. The `encode_for_hash` function can also substitute the `signature_to_octets` and `proof_to_octets` operations (the results will be the same). I left that for another PR if we decide to go that way (I’m not convinced that it would be the best option myself).
2. Removed the distinction with DSTs. We are explicitly appending their length everywhere in the spec.
3. this pr also closes #74 like pr #188. However, I realize this requires more time and reviews so better to move with #188 first (I will resolve the conflicts).


BasileiosKal:
Split the challenge hash input to multiple lines.

draft-bbs-signatures.md
```diff
@@ -499,17 +538,19 @@ Procedure:
...
 5. if W is INVALID, return INVALID
 
-6. gen_octets = (H_s || H_d || H_1 || ... || H_L)
+6. dom_array = (PK, L, H_s, H_d, H_1, ..., H_L, Ciphersuite_ID, header)
 
-7. domain_prime = (PK || L || gen_octets || Ciphersuite_ID || header)
+7. dom_for_hash = encode_for_hash(dom_array)
 
-8. domain = hash_to_scalar(domain_prime, 1)
+8. if dom_for_hash is INVALID, return INVALD
```
andrewwhitehead, line 545:
```suggestion
8. if dom_for_hash is INVALID, return INVALID
```

draft-bbs-signatures.md
```diff
@@ -711,27 +760,35 @@ Procedure:
 
 5. if W is INVALID, return INVALID
 
-6. gen_octets =  (H_s || H_d || H_1 || ... || H_L)
+6. dom_array = (PK, L, H_s, H_d, H_1, ..., H_L, Ciphersuite_ID, header)
+
+7. dom_for_hash = encode_for_hash(dom_array)
+
+8. if dom_for_hash is INVALID, return INVALD
```
andrewwhitehead, line 767:
```suggestion
8. if dom_for_hash is INVALID, return INVALID
```

draft-bbs-signatures.md
```diff
@@ -585,47 +626,55 @@ Procedure:
 
 3. (A, e, s) = signature_result
 
-4. gen_octets = (H_s || H_d || H_1 || ... || H_L)
+4. dom_array = (PK, L, H_s, H_d, H_1, ..., H_L, Ciphersuite_ID, header)
+
+5. dom_for_hash = encode_for_hash(dom_array)
+
+6. if dom_for_hash is INVALID, return INVALD
```
andrewwhitehead, line 633:
```suggestion
6. if dom_for_hash is INVALID, return INVALID
```

BasileiosKal:
Ready for review c.c. @andrewwhitehead, @tplooker ^

draft-bbs-signatures.md
```diff
@@ -427,21 +460,27 @@ Precomputations:
...
+2. dom_for_hash = encode_for_hash(dom_array)
 
-3. domain = hash_to_scalar(domain_prime, 1)
+3. if dom_for_hash is INVALID, return INVALID
 
-4. (e, s) = hash_to_scalar((SK  || domain || msg_1 || ... || msg_L), 2)
+4. domain = hash_to_scalar(dom_for_hash, 1)
 
-5. B = P1 + H_s * s + H_d * domain + H_1 * msg_1 + ... + H_L * msg_L
+5. e_s_for_hash = encode_for_hash(SK, domain, msg_1, ..., msg_L)
```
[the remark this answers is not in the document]
BasileiosKal: Good catch! Thanks!

andrewwhitehead (approved):

draft-bbs-signatures.md
```diff
@@ -255,34 +258,67 @@ Aside from the message generators, the scheme uses two additional generators: `H
 
 ## Encoding of elements to be hashed
```
tplooker, line 259: Should we shift this as an operation under section 5 (https://identity.foundation/bbs-signature/draft-bbs-signatures.html#name-utility-operations)?

Probably best since we are referencing `hash_to_scalar` also!

BasileiosKal:
After @tplooker's suggestion, I made the procedure "informative" (made it a recommendation) and added the "normative" definition to the notation section (maybe we will need to do that for `hash_to_scalar` and `create_generators` as well).

@andrewwhitehead really sorry to requesting your review again. Any thoughts on that??^

andrewwhitehead:
I don't think it can be only a recommended implementation, as any changes would impact compatibility and the security of the scheme. If variations were supported then I think they would need to be added to the suite identifier.

BasileiosKal:
> I don't think it can be only a recommended implementation, as any changes would impact compatibility and the security of the scheme. If variations were supported then I think they would need to be added to the suite identifier.

@andrewwhitehead that is true. The tradeoff IMO is that if the `encode_for_hash` is "normative" we may need to define it more formally. However this will be hard to do in the spec (for example is the line `if el is an octet string representing a public key` formal enough??). I'm afraid that if we make it a requirement with this type of syntax we may face the same negative feedback that the [ZCach encoding](https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-10.html#name-zcash-serialization-format-) in the [PFC spec](https://www.ietf.org/archive/id/draft-irtf-cfrg-pairing-friendly-curves-10.html) faced by the CFRG.

Don't have a perfect solution for now. Should we leave it as normative and revisit this later (i.e., not close #185)?? Thoughts @tplooker??^

tplooker:
Apologies I should have been more specific, I think the behaviour of encode_for_hash needs to be normative to promote interoperable implementations as @andrewwhitehead points out. I was more wondering whether we position the procedural definition here as informative because I can imagine there being multiple ways to implement it in practise. I guess that argument could also be applied more broadly to other operations in the scheme too so perhaps its not worth the complication?

BasileiosKal:
Updated. As far as if the procedure is "formal enough" to be normative, I'm not 100% convinced my self but I may be overthinking this. 

It's true that most likely people will not use a straight line implementation of this procedure, however i do mention that this is just an explanation of the described behavior, and also most likely, to @tplooker's point, people will not use straight line implementations for most things in the spec IMO.

draft-bbs-signatures.md
```diff
@@ -1146,6 +1135,70 @@ Procedure:
...
+- ASCII strings will be transformed into octet strings using UTF-8 encoding.
+
+After encoding, octet strings will be prepended with a value representing the length of their binary representation in the form of the number of bytes. This length must be encoded to octets using I2OSP with output length of 8 bytes. The combined value (encoded value + length prefix) binary representation is then encoded as a single octet string. For example, the string `0x14d` will be encoded as `0x0000000000000002014d`. If the length of the octet string is larger than 2^64 - 1, the octet string must be rejected. Similarly, ASCII strings, after encoded to octets (using utf8), will also be appended with the length of their octet-string representation.
+
+Optional input/parameters to operations that feature in a call to hash\_to\_scalar, that are not supplied to the operation should default to an empty octet string. For example, if X is an optional input/parameter that is not supplied, whilst A and B are required, then the procedural step of `hash(A || X || B)` MUST be evaluated to `hash(A || "" || B)`.
+
+The above procedure is further described in the following operation.
```
tplooker, line 1152:
```suggestion
The above is further described in the following procedure.
```

draft-bbs-signatures.md
```diff
@@ -1146,6 +1135,70 @@ Procedure:
 5. return W
 ```
 
+### EncodingForHash
```
tplooker, line 1138:
```suggestion
### EncodeForHash
```

tplooker (approved):

tplooker:
Multiple approvals, discussed on 4th of July WG call, merging

PR #199, opened 2022-07-06
consistency updates

BasileiosKal:
Updates discussed in the last (on 2022-07-04) WG call:

1. `H_s`, `H_d` -> `Q_1`, `Q_2`
2. `messageIndexes`, `disclosedMessages` -> `message_indexes`, `disclosed_messages`
3. `Ciphersuite__ID` -> `ciphersuite_id`
4. Added `length` definition
5. Removed the `H_s` and `H_d` definitions

Some of the updates that i didn't make:
1. Didn’t change `MsgGenerators` to `msg_generators` (it looks better this way to me, but this is the only reason. Happy to make the update if people prefer it)
2. Didn’t update `P1` to `Q_1`. I’m worried about the security of that change in some scenarios, and I think we should think a little more about it. Will open an issue to discuss more and explain my worries.

### DIFF
--- a/draft-bbs-signatures.md
+++ b/draft-bbs-signatures.md
@@ -145,12 +145,6 @@ msg
 generator
 : A valid point on the selected subgroup of the curve being used that is employed to commit a value.
 
-H_s
-: A generator for the blinding value in the signature. The value of H_s is defined by each ciphersuite and must always be supplied to the operations listing it as a parameter.
-
-H_d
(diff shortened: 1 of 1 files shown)


andrewwhitehead (approved):

tplooker (approved):

tplooker:
Multiple approvals, proposal discussed on WG call, merging.

PR #202, opened 2022-07-08
length definition fix

BasileiosKal:
Small fix in the definition of length (length is used for both arrays and octet strings).


tplooker (approved):

tplooker:
Minor clarification, merging

PR #205, opened 2022-08-17
Removed generator_seed from create_generators calls

christianpaquin:
Removed the `generator_seed` parameter from the `create_generators` calls, as it is already defined as a ciphersuite value within that function.


tplooker (approved):

BasileiosKal (approved):

BasileiosKal:
Multiple approvals! Editorial update. Merging! Thanks @christianpaquin!!

PR #208, opened 2022-08-18
Update signature related fixtures

tplooker:
This PR updates the fixtures to align to the current spec, including adding in those that feature a header value.

PR #210, opened 2022-08-18
Update proof fixtures

tplooker:
Depended on #208 being merged ahead


draft-looker-cfrg-bbs-signatures.md
```diff
@@ -1259,6 +1271,156 @@ Along with the SK value as defined in (#key-pair) as inputs into the Sign operat
...
+{{ $proofFixtures.proof002.presentationMessage }}
+```
+
+And public key of the signer
+
+```
+{{ $proofFixtures.proof002.signerPublicKey }}
+```
+
+Yields the following proof
```
BasileiosKal, line 1362: Maybe it's naive, but could this be confusing until we have the ability to test with a deterministic proofGen??

christianpaquin:
Question about the values used to generate these test vectors: do all the signatures and proofs use the generators in `fixtures/generators.json`, and if so, are the two first entries `Q1` and `Q2` and the remaining ones `H1` ... `H8` ?

tplooker:
DIscussed on WG call 20th Sept, will add a note to capture the fixtures are non-deterministic in nature.

christianpaquin:
It would be great to also add the fixtures for the sha256 ciphersuite

christianpaquin:
> It would be great to also add the fixtures for the sha256 ciphersuite

Any ETA for this, @tplooker, @BasileiosKal? Anxious to plug these in my implementation :)

BasileiosKal:
Hey @christianpaquin! Sorry for the long delay. Had to take some time off. Iv now opened #221 with the updated fixtures. 

Note that the signature fixtures do also take into account #216. It's a simple update were the ciphersuite id is not appended with its length when hashing.

tplooker:
Closing in favour of #221

PR #211, opened 2022-08-19
Add DST as input to hash_to_scalar

### LINKED ISSUE #194: expand_message domain separation
The current hash_to_scalar construction does not allow for the calling operation to define the DST that will go into the underlying expand_message used. Some operations that need additional domain separation append the DST to the message.

The problem is that we cannot prove indistinguishably that way. For example consider the following, lets assume that in a protocol we call hash_to_scalar in the input `msg ||  I2OSP(length(DST2), 1) || DST2` and get `b_1, b_2, ...` where,

```
b_0 = H(Z_pad || msg ||  I2OSP(length(DST2), 1) || DST2 || I2OSP(len_in_bytes, 2) || I2OSP(0, 1) || DST_prime)
b_1 = H(b_0 || ...)
b_2 = ....
```
Then, in some other place, the protocol returns `h2 = H(Z_pad || msg || I2OSP(length(DST2), 1) || DST2)`. In that case, after observing `h2` I could do a length-extension attack on H, calculate `b_0` and then `b_1`, `b_2` etc. This is not really a practical issue (it becomes less practical -but not impossible- if we also consider padding), but it does stop us from proving security. It also becomes a larger issue if people use a different expand_message (if they just use sha512 with a counter for example).

### Solution A
The currently proposed solution, discussed in the WG call, would be to define an additional optional argument to hash_to_scalar like `dstInfo` and append that to the DST already defined. i.e., inside hash_to_scalar calculate,
```
4. dst_prime = DST || dstInfo || I2OSP(length(dstInfo), 1)
5. expand_message(msg, dst_prime, len_in_bytes)
```
This is a bit weird because we are prepending the length everywhere else in the spec. However, if we do the same here and calculate `dst_prime = DST || I2SOP(length(dstIfo), 1) || dstInfo` then for example the pairs `DST1 = [4, 2]`, `dstInfo1 = [2]` and `DST2 = [4]`, `dstInfo2 = [1, 2]` will give the same `dst_prime`. 

Similarly, we can't just concatenate the 2 (i.e., `dst_prime = DST || dstInfo`) because then for example the pairs `DST1 = [1, 2]`, `dstInfo1 = [3]` and `DST2 = [1]`, `dstInfo2 = [2, 3]` will again give the same `dst_prime`

### Solution B
Don't define the DST in hash_to_scalar but require it as an input. So whatever operation calls hash_to_scalar will also need to specify a DST value to pass it. This is probably the most efficient approach when it comes to a straight line implementation, but IMO the "ugliest" looking when it comes to the spec.

BasileiosKal:
Closes #194

I also updated the definition of the `ciphersuite_id` to be a ASCII string, instead of an octet string. This makes it easier to concatenate it with other strings to create DSTs which will then become octet strings using `utf8()`.

Lastly, I'm passing the inputted message in `MapMessageToScalarAsHash` directly to `hash_to_scalar` because I don't see a reason to append it with each length anymore, since we are using `expand_message`

### DIFF
--- a/draft-looker-cfrg-bbs-signatures.md
+++ b/draft-looker-cfrg-bbs-signatures.md
@@ -705,10 +705,9 @@ result = MapMessageToScalarAsHash(msg, dst)
 Inputs:
 
 - msg (REQUIRED), octet string.
-- dst (OPTIONAL), an octet string representing an ASCII encoded domain
-                  separation tag. If not supplied, it defaults to the
-                  octet string given by
-                  utf8(ciphersuite_id || "MAP_MSG_TO_SCALAR_AS_HASH_"),
+- dst (OPTIONAL), an octet string representing a domain separation tag.
+                  If not supplied, it default to the octet string
+                  utf8(ciphersuite_id || "MAP_MSG_TO_SCALAR_AS_HASH_")
                   where ciphersuite_id is defined by the ciphersuite.
 
 Outputs:
@@ -739,10 +738,10 @@ Inputs:
 - msg_octets (REQUIRED), octet string. The message to be hashed.
 - count (REQUIRED), an integer greater or equal to 1. The number of
                     scalars to output.
-- dst (OPTIONAL), an octet string, representing an ASCII encoded domain
-                  separation tag. If not supplied, it defaults to the
-                  octet string given by utf8(ciphersuite_id || "H2S_"),
-                  where ciphersuite_id is defined by the ciphersuite.
+- dst (OPTIONAL), an octet string representing a domain separation tag.
+                  If not supplied, it defaults to the octet string given
+                  by utf8(ciphersuite_id || "H2S_"), where
+                  ciphersuite_id is defined by the ciphersuite.
 
 Parameters:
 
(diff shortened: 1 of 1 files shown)


draft-looker-cfrg-bbs-signatures.md
```diff
@@ -697,45 +697,52 @@ There are multiple ways in which messages can be mapped to their respective scal
...
 
 - msg (REQUIRED), octet string.
-- dst (REQUIRED), octet string. Domain separation tag; note this is not
-                  defined as a function argument as per
-                  [@!I-D.irtf-cfrg-hash-to-curve] but as a parameter.
+- dst (OPTIONAL), an octet string representing an ASCII encoded domain
+                  separation tag. If not supplied, it defaults to the
+                  octet string given by
+                  utf8(ciphersuite_id || "MAP_MSG_TO_SCALAR_AS_HASH_"),
+                  where ciphersuite_id is defined by the ciphersuite.
```
andrewwhitehead, line 712: I don't think the DST needs to be defined as 'ASCII encoded' here, as technically any octet string is accepted.

tplooker:
Discussed on WG call 29th August, awaiting updates and re-review.

draft-looker-cfrg-bbs-signatures.md
BasileiosKal: Done. Thanks!

andrewwhitehead (approved):

tplooker (approved):

tplooker:

PR #214, opened 2022-09-12
Generator fixtures update

BasileiosKal:
Updated generator fixtures. Also disabled the fixture-population workflow until the tooling is up to date with the spec.

Issue #206, opened 2022-08-18
create_generators flexibility


BasileiosKal:
Currently we allow seeds with different scopes but it is not clear how a signature specific seed is going to be used since it is passed as a parameter to `create_generators` and not as input. Also i think this is a case that we want to reduce optionality (for example we don't want `generator_seed = hash(all_messages)`).

I think it's best to not allow "scoped seeds" but different `create_generators` operations. Each create_generators will accept a `seed` and a `dst` as a parameter and a `count` as input. If one wants signature specific generators they can use the `count` input but nothing else. This will be enough for "legacy support". 

This also makes it easier to describe the requirements of the `create_generators` operations (rather than the requirements for the seed selection).

IMO there are 2 options:
1. Don't allow different `create_generators` or `scoped_seeds`: No changes are needed, expect to remove the text around "scoped seeds".
2. Allow different `create_generators`: In this case IMO we should:
    - Add a `create_generators_operation` parameter in the ciphersuites definition.
    - For the suites defined in the document give the `create_generators_operation` parameter the value of "`create_generators` with `generator_seed=...`".
    - Require every create_generators operation to also define a unique ID which will be added to the ciphersuite_id (like EXP_TAG for expand_message in [h2c](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#name-suite-id-naming-conventions)).
    - Add `create_generators_operation` as a parameter to Sign, Verify etc..
    - Add the requirements for defining a new `create_generators_operation`.

Obviously 1 is the simplest but the least flexible. It also is a "breaking change" with previous versions. So my preference will be 2 but I'm also very interested in others thoughts on the matter.

BasileiosKal:
Discussed on last WG call. Will open a PR to the proposed direction.

PR #215, opened 2022-09-12
Add suite specific fixed g1 base point

BasileiosKal:
Updated tooling to calculate the G1 base point for each ciphersuite. 

- To avoid conflicts with #214 I don't write them in the fixtures file (the points are hardcoded in the suites). Will update after that PR is resolved.
- The message generator fixtures creation still needs updating to the latest spec.

### DIFF
--- a/draft-looker-cfrg-bbs-signatures.md
+++ b/draft-looker-cfrg-bbs-signatures.md
@@ -363,6 +363,7 @@ Inputs:
 Parameters:
 
 - ciphersuite_id, ASCII string. The unique ID of the ciphersuite.
+- P1, fixed point of G1, defined by the ciphersuite.
 
 Definitions:
 
@@ -418,6 +419,7 @@ Inputs:
 Parameters:
 
 - ciphersuite_id, ASCII string. The unique ID of the ciphersuite.
+- P1, fixed point of G1, defined by the ciphersuite.
 
 Definitions:
 
@@ -480,6 +482,7 @@ Inputs:
 Parameters:
 
 - ciphersuite_id, ASCII string. The unique ID of the ciphersuite.
+- P1, fixed point of G1, defined by the ciphersuite.
 
 Definitions:
 
@@ -574,6 +577,7 @@ Inputs:
 Parameters:
 
 - ciphersuite_id, ASCII string. The unique ID of the ciphersuite.
+- P1, fixed point of G1, defined by the ciphersuite.
 
 Definitions:
 
@@ -653,6 +657,7 @@ Parameters:
                   specified by the hash_to_curve_suite parameter.
 - generator_seed, octet string. A seed value selected by the
                   ciphersuite.
+- P1, fixed point of G1, defined by the ciphersuite.
 
 Definitions:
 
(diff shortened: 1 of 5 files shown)


OR13 (approved):

draft-looker-cfrg-bbs-signatures.md
```diff
@@ -677,7 +682,7 @@ Procedure:
 5.     n = n + 1
 6.     generator_i = Identity_G1
 7.     candidate = hash_to_curve_g1(v, generator_dst)
-8.     if candidate in (P1, generator_1, ..., generator_i):
+8.     if candidate in (generator_1, ..., generator_i):
```
jovfer, line 685: @BasileiosKal @OR13  most likely a dumb question(s) from me, as I'm not an expert in BBS+:
1) Is it OK to have a collision for some generator_i == P1?
2) Or is it guaranteed that hash_to_curve_g1 <> P1? (why?)

may be it's worth a note here.

OR13: cc @mikelodder7 @tplooker

BasileiosKal: That’s actually a great question😄

## TL;TR
It's not OK to have a collision, but the probability of having a collision with P1 is practically 0.

## In detail
That probability is ~2^-128. This comes from the collision resistance of hash_to_curve_g1 (which comes from the collision resistance of the underlying hash function+the security level of the curve). This guarantees that unless you create more than 2^128 generators, generator_i will be different from P1 (note that they are created from different inputs to `hash_to_curve_g1`).

If that security level is deemed not enough in the future, there are alternative curves and hash functions that can be used, with higher security levels.

We have checks like that throughout the document, but we are currently debating if we should keep all of them. The tradeoff is that they guard against events that could compromise security, however, the probability of those events is practically 0. Also, we are worried that the more complicated draft could lead to implementation bugs like infinite loops (since we keep trying in a loop until we get a valid result).

A note is a good idea. We do have an [issue](https://github.com/decentralized-identity/bbs-signature/issues/206) open, to debate on what the best way to describe those requirements would be. No matter the case though, we will soon add text that describes all those considerations in detail.

mikelodder7: Agreed. The likelihood of a collision is so low it’s not worth checking. Just a note about it is good enough

jovfer: Thanks for the details, that makes perfect sense. My understanding was similar, but the next things is actually the source for my confusion: 
What about of the collusion for `generator_i` and `generator_j`? isn't it the same to `P1` vs `generator_i` ?

Should we remove the whole check in this line?

BasileiosKal: Collisions between `generator_i` and `generator_j` are a bit different. Because we are essentially checking for (any) duplicates (rather than equality with a fixed value, i.e., `generator_i = P1`), we must also consider the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_problem) when calculating the probability of that event.
 
That said, it probably won’t matter enough in practice, and the probability will still be close to ~2^-128. So most likely we will remove that check too. It’s a bit more complicated case though so we want to be 100% sure first.

jovfer: ah yes, I've missed the effect of the birthday paradox. Not sure how much actual effect it will have for avg number of msgs per credential for such small probability per single collision, but agree: let's keep the rest of the check for now.
Thanks for the explanation.

andrewwhitehead (approved):

tplooker (approved):

PR #216, opened 2022-09-23
Update encoding rules

### LINKED ISSUE #207: ascii and utf8
I don't think we are consistent about this. For example we define utf8 as accepting only ascii strings and we don't handle non ascii strings in `encode_for_hash`, even though we have an utf8 encoding. Also, I think "ASCII character sequence" is what we meant as "ascii string" but is not everywhere clear. 

Not sure what is best here. Should we assume that all the messages are already an octet string (and just mention some ways for transforming a char sequence to octets)?? So the only think that is left is DST's that we can encode with utf8??

I think this is what most crypto specs do, so i would go with that.

BasileiosKal:
Closes #207

DSTs are defined as octet strings everywhere. Default DSTs are turned into octet strings using `utf8`. Also, the `ciphersuite_id` is excluded from the “must be prepended with its length” rule when hashing (no reason for it).


draft-looker-cfrg-bbs-signatures.md
```diff
@@ -1002,12 +1005,15 @@ Procedure:
 1.  let octets_to_hash be an empty octet string.
 2.  for el in input_array:
 3.      if el is an ASCII string: el = utf8(el)
-4.      if el is an octet string representing a public key: el_octs = el
-5.      else if el is an octet string:
-6.          if length(el) > 2^64 - 1, return INVALID
-7.          el_octs = I2OSP(length(el), 8) || el
-8.      else if el is a Point in G1: el_octs = point_to_octets_g1(el)
-9.      else if el is a Point in G2: el_octs = point_to_octets_g2(el)
+4.      if el is an octet string representing a public key:
```
andrewwhitehead, line 1008: Is this used? Why not just pass it as a G2 point?

draft-looker-cfrg-bbs-signatures.md
```diff
@@ -656,11 +656,12 @@ Parameters:
 
 Definitions:
 
-- seed_dst, the octet string representing the ASCII encoded characters:
-           "BBS_" || hash_to_curve_suite || "SIG_GENERATOR_SEED_".
-- generator_dst, the octet string representing:
-                 "BBS_" || hash_to_curve_suite || "SIG_GENERATOR_DST_",
-                 in the ASCII characters encoding.
+- seed_dst, octet string representing the domain seperation tag:
+            urtf8(ciphersuite_id || "SIG_GENERATOR_SEED_"), where
```
andrewwhitehead, line 660:
```suggestion
            utf8(ciphersuite_id || "SIG_GENERATOR_SEED_"), where
```

tplooker:
With the CFRG draft adoption the draft name has changed again, please rebase this PR to target the new draft name `draft-irtf-cfrg-bbs-signatures`

draft-looker-cfrg-bbs-signatures.md
```diff
@@ -1002,12 +1005,15 @@ Procedure:
...
-4.      if el is an octet string representing a public key: el_octs = el
-5.      else if el is an octet string:
-6.          if length(el) > 2^64 - 1, return INVALID
-7.          el_octs = I2OSP(length(el), 8) || el
-8.      else if el is a Point in G1: el_octs = point_to_octets_g1(el)
-9.      else if el is a Point in G2: el_octs = point_to_octets_g2(el)
+4.      if el is an octet string representing a public key:
+5.          el_octs = el
+6.      else if el is representing a utf8 encoded Ciphersuite ID:
+7.          el_octs = utf8(el)
```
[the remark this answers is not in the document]
BasileiosKal: Thinks! Fixed.

draft-looker-cfrg-bbs-signatures.md
BasileiosKal: Thanks! Fixed.

draft-looker-cfrg-bbs-signatures.md
BasileiosKal: We definitely could, but i like having the PK as an octet string. It's more intuitive for a developer IMO. Also updating this would be a bigger change (will need to update the `octets_to_pubkey` to something like `validate_pubkey` which I'm not sure how it would look like. Opened #219 to keep track of this.

tplooker (approved):

andrewwhitehead (approved):

tplooker:
Multiple approvals, open over two weeks, merging

PR #220, opened 2022-10-07
Tooling: fix DST for h2c ops

jovfer:
Align msg generator tooling to spec.

https://github.com/decentralized-identity/bbs-signature/blob/main/draft-irtf-cfrg-bbs-signatures.md#generator-point-computation - step 7th of 'Procedure'

### DIFF
--- a/tooling/message-generators/src/main.rs
+++ b/tooling/message-generators/src/main.rs
@@ -172,7 +172,7 @@ where
     let mut buffer = [0u8; 64];
     for _ in 0..len {
         reader.read(&mut buffer);
-        let gi = G1Projective::hash::<ExpandMsgXof<Shake256>>(&buffer, DST);
+        let gi = G1Projective::hash::<ExpandMsgXof<Shake256>>(&buffer, &X::generator_dst());
         generators.push(gi);
     }
 


tooling/message-generators/src/main.rs
```diff
@@ -172,7 +172,7 @@ where
     let mut buffer = [0u8; 64];
     for _ in 0..len {
         reader.read(&mut buffer);
```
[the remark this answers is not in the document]
BasileiosKal: Its a create_generators procedure from a previous version of the spec and should be updated. Thanks for pointing it out!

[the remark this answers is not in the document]
jovfer: done

tooling/message-generators/src/main.rs
```diff
@@ -158,22 +158,27 @@ fn make_generators<'a, X>(seed: Option<&[u8]>, len: usize) -> Generators
 where
     X: BbsCiphersuite<'a>
 {
-
+    const EXPAND_LEN: usize = 48; //TODO make a property of X or calculate?
     let default_seed = &X::generator_seed();
-    let seed = seed.unwrap_or(default_seed);
+    let seed = seed.unwrap_or(default_seed); //TODO remove parameter and always use default one?
```
BasileiosKal, line 163:
```suggestion
    let seed = seed.unwrap_or(default_seed);
```
This is just regarding the TODO ;P In the spec there is the option of supplying an issuer specific seed and not the default one. There is some discussion around this and it may change in the future, but currently this is here to allow for this (note that the `signer_specific_generators` will pass the Issuer's pk as the seed).

tooling/message-generators/src/main.rs
```diff
@@ -158,22 +158,27 @@ fn make_generators<'a, X>(seed: Option<&[u8]>, len: usize) -> Generators
...
+    let mut buffer = vec!(0u8; EXPAND_LEN);
+    X::Expander::expand_message(seed, &X::generator_seed_dst(), &mut v);
+
+    let mut n = 1u32;
+    while generators.len() < len {
+        v.append(n.to_be_bytes().to_vec().as_mut());
+        X::Expander::expand_message(&v, &X::generator_seed_dst(), &mut buffer);
+        v = buffer.clone();
+        n += 1;
+        let candidate = G1Projective::hash::<ExpandMsgXof<Shake256>>(&v, &X::generator_dst());
```
BasileiosKal, line 178:
```suggestion
        let candidate = G1Projective::hash::<<X as BbsCiphersuite>::Expander>(&v, &X::generator_dst());
```
This is essentially the hash_to_cuurve_g1 call in the procedure. Hash to curve should use whatever expand_message operation the ciphersuite defines (its actually the other way around. The ciphersuite defines a hash-to-curve suite which in turn defines the expand_message used, but you get the point 😄).

BasileiosKal (commented):
Great! Thanks for working on this! Just a couple of comments. LGTM otherwise!

jovfer:
@BasileiosKal thanks for review :) applied your suggestions

tooling/message-generators/src/main.rs
jovfer: oops, fixed one hard-coded thing (DST) but missed the main one 😄

BasileiosKal (approved):
LGTM! Thanks!

tplooker (approved):

tplooker:
Discussed on WG call 17th of October, multiple approvals, merging thanks @jovfer!

PR #221, opened 2022-10-08
Fixtures and tooling update

### LINKED ISSUE #217: P1 calculation in tooling code
As per spec, P1 is calculated using `create_generators` with a specific `generator_seed`. `create_generators` procedure as described [here](https://identity.foundation/bbs-signature/draft-looker-cfrg-bbs-signatures.html#name-generator-point-computation) starts calculation with `n=1` but below reference reference code in `tooling`  to calculate P1 uses `n=0`. 
https://github.com/decentralized-identity/bbs-signature/blob/3e69a8b16fd7788993e9d3a8849f311ade47fc5e/tooling/message-generators/src/main.rs#L193
I think the code should be updated as below and then the new values of `P1` also need to be updated in the cipher suites constants section.
```rust 
let extra = 1usize.to_be_bytes()[4..].to_vec();
```

BasileiosKal:
This PR:
- Updates existing fixtures and adds SHA256 suite ones. Proof fixture files are included but not added to the spec to avoid further conflicts with #210.
- Updates the fixtures tooling to read and write the fixtures on the document.
- Closes #217. Updates the message-generators tooling to return the correct file format.

## Notes 

1. This PR depends on #216 (in all the fixtures i am not prepending the length of the ciphersuite's ID encoding when hashing).
2. The message-generators tooling is "deactivated" on the workflow since it is not up to date with the spec. The generator files are not the ones returned by that tool and should be the correct ones.


tmarkovski:
Would it be possible to include fixtures for some of the other utility operations, specifically `scalars = hash_to_scalar(msg_octets, count)`. This operation is harder to validate as it's always a dependent part inside existing operations. A simple fixture for hashing an octet string into 1 and 2 scalars would go a long way.

tooling/fixtures/fixture_data/bls12-381-sha-256/signature/signature001.json
```diff
@@ -8,7 +8,7 @@
   "messages": [
     "9872ad089e452c7b6e283dfac2a80d58e8d0ff71cc4d5e310a1debdda4a45f02"
```
tmarkovski, line 9: Is this item a scalar value? If so, is it in big endian?

tooling/fixtures/fixture_data/messages.json
```diff
@@ -0,0 +1,12 @@
+[
+    "9872ad089e452c7b6e283dfac2a80d58e8d0ff71cc4d5e310a1debdda4a45f02",
```
tmarkovski, line 2: Are these messages treated as scalars that can be direct inputs in the `sign` operation? If so, I have issues deserializing some of these as scalar values. I may be doing something wrong, but would love some clarification how we use these values.

BasileiosKal: They need to be mapped using `MapMessageToScalarAsHash` prior to passing them in Sign etc. Thanks for pointing out this oversight! Will add the scalar values as well as some `hash_to_sclar` fixtures soon.

tmarkovski: Much appreciated

tmarkovski (commented):
Calls in `main.rs` in `make_generators` to `G1Projective::hash::<ExpandMsgXof<Shake256>>` should be replaced with `G1Projective::hash::<X::Expander>`. Sorry, I can't make a suggestion, as it's not part of the diff changes.

tmarkovski:
Some of the DST or SEED values contain `RO_` in them, while some use `R0_`, letter `O` vs number `0`. Is this intentional?

BasileiosKal:
R.e., the `R0_` in the DST's it's a typo. Thanks for identifying it, would never had picked it up 😅 The correct form is `RO_` (not with the zero). The fixtures use the correct DST (with `RO_`)

R.e., the `make_generators` procedure, the current one is not up to date with the latest spec. PR #220 updates it. This PR has "deactivated" that tooling for the workflow, i.e., the generators fixtures are not yet created from that tool. Will reactivate it after #220 is merged.

BasileiosKal:
Added `hash-to-scalar` and `MapMessageToScalarAsHash` test vectors. @tmarkovski hope those will be useful!

tplooker (approved):

tmarkovski (approved):

.github/workflows/ghpages.yml
```diff
@@ -21,23 +21,23 @@ jobs:
     steps:
       - name: "Checkout"
         uses: actions/checkout@v2
+        
+      # Temporarily disabled until the tool is updated to the latest spec
```
jovfer, line 25: Do we need more updates in the tool?

BasileiosKal:
@tmarkovski, @christianpaquin I just pushed some new test vectors, after fixing a typo i made in one of the dst's. Sorry about that. It should be consistent now. Please use the updated ones in your testings!

tmarkovski:
Happy to report that all test vectors are passing at present! Thanks for updating the test vectors and looking into the my issue.

<details>
  <summary>Test Results</summary>

```
running 54 tests
test hashing::test::test_encode ... ok
test encoding::test::encode_for_hash_test_vector ... ok
test encoding::test::to_octet_string_test ... ok
test hashing::test::hash_to_curve_1_scalar_output_shake ... ok
test hashing::test::hash_to_curve_1_scalar_output_sha ... ok
test hashing::test::hash_to_curve_10_scalar_output_shake ... ok
test hashing::test::map_message_to_scalar_test ... ok
test key::test::get_random_key ... ok
test key::test::gen_key_from_ikm ... ok
test key::test::sk_to_pk_test ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof008_json ... ok
test generators::test::generators_test::case_bls12_381_sha_256_generators_json_bls12381sha256 ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof007_json ... ok
test generators::test::generators_test::case_bls12_381_shake_256_generators_json_bls12381shake256 ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof009_json ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof001_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof001_json ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof004_json ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof005_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof007_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof008_json ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof003_json ... ok
test signature::test::signature_from_octets_fails_incorrect_size ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof006_json ... ok
test signature::test::signature_from_octets_succeeds_slice ... ok
test signature::test::signature_from_octets_succeeds ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof009_json ... ok
test proof::test::test_proof_from_bytes ... ok
test proof::test::proof_suite_1::case_bls12_381_sha_256_proof_proof002_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof002_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof004_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof003_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof005_json ... ok
test proof::test::proof_suite_2::case_bls12_381_shake_256_proof_proof006_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature003_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature002_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature001_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature005_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature004_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature006_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature007_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature005_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature003_json ... ok
test signature::test::signature_to_octets_succeeds ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature002_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature001_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature008_json ... ok
test signature::test::signature_suite_1::case_bls12_381_sha_256_signature_signature009_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature004_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature006_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature007_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature008_json ... ok
test signature::test::signature_suite_2::case_bls12_381_shake_256_signature_signature009_json ... ok
test test::bbs_demo ... ok

test result: ok. 54 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.81s


running 5 tests
test src/lib.rs - Bbs::message (line 65) ... ok
test src/lib.rs - Bbs::message_with (line 89) ... ok
test src/lib.rs - Bbs::message (line 56) ... ok
test src/lib.rs - Bbs::sign (line 115) ... ok
test src/lib.rs - Bbs::verify (line 143) ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.58s
```

</details>

My thoughts and findings:
- Most negative tests are hard to validate for the reason they specify. Once my positive tests worked, I only assumed the negative tests are working as well, since they were all failing before.
- Tests without header and presentation header might be useful as well
- I stumbled on the use of hex and treating messages as encoded scalars, but this has already been addressed I think. Might be useful to add variable length message inputs.
- Why does the `Sign` algorithm accept SK and PK as input, but doesn't validate if the PK is correct. `SkToPk` is fairly inexpensive operation, it seems that only SK is needed. Is there a security concern similar to what affected ECDSA?

tooling/message-generators/src/main.rs
```diff
@@ -190,7 +222,7 @@ where
     X::Expander::expand_message(&X::bp_generator_seed(), &X::generator_seed_dst(), &mut v);
 
     // TODO: implement a proper I2OSP
-    let extra = 0usize.to_be_bytes()[4..].to_vec();
+    let extra = 1usize.to_be_bytes()[4..].to_vec();
```
tmarkovski, line 225:
```suggestion
    let extra = 1u32.to_be_bytes().to_vec();
```

This will fail on 32-bit systems. Using any other fixed length integer is a better approach.

jovfer:
@tmarkovski is the implementation and tests (referred in your message above) public? If so, could you please share a link?

christianpaquin:
It would be good to add a signing test over 0 messages, and a proof test disclosing no messages, to provide test cases for edge conditions (to validate the handling of empty argument arrays).

tooling/fixtures/fixture_data/bls12-381-sha-256/signature/signature001.json
christianpaquin, line 9: How are these messages encoded, @BasileiosKal? Are these scalars (in hex), or octet strings that need to be encoded with MapMessageToScalar? When I try to parse these directly, I get an error because these are bigger than the order `r`.

christianpaquin: Looking at @tmarkovski's implementation, seems like you indeed need to first map the messages to scalar. This is a bit confusing because the spec's Sign and Verify function's `messages` param, matching the name in the fixture files, is of type scalar.

tplooker: @christianpaquin agreed I think clarifying text here about the form of messages in these test vectors is important. Also think the additional test case you suggest is a good one, there are few possible ones we could add.

tplooker:
> @tmarkovski are the implementation and tests (referred in your message above) public? If so, could you please share a link?

I believe it is this @jovfer? https://github.com/trinsic-id/bbs/tree/main

tooling/fixtures/fixture_data/bls12-381-sha-256/proof/proof001.json
```diff
@@ -0,0 +1,14 @@
+{
+  "caseName": "single message signature, message revealed proof",
+  "signerPublicKey": "b65b7cbff4e81b723456a13936b6bcc77a078bf6291765f3ae13170072249dd7daa7ec1bd82b818ab60198030b45b8fa159c155fc3841a9ad4045e37161c9f0d9a4f361b93cfdc67d365f3be1a398e56aa173d7a55e01b4a8dd2494e7fb90da7",
+  "header": "11223344556677889900aabbccddeeff",
+  "presentationHeader": "bed231d880675ed101ead304512e043ade9958dd0241ea70b4b3957fba941501",
+  "revealedMessages": {
+    "0": "9872ad089e452c7b6e283dfac2a80d58e8d0ff71cc4d5e310a1debdda4a45f02"
```
christianpaquin, line 7: All the revealed messages are numbered using a 0-based counter in the tests. My implementation uses a 1-based one, matching the sub-indices in the spec. This could have interop implications when these proofs are serialized, so we should call out the expected behavior. My personal vote: use a one-based counter.

christianpaquin:
FYI @BasileiosKal, my [implementation](https://github.com/microsoft/bbs-node-reference) now validates these test vectors; simply sync with main and `npm test`.

tooling/fixtures/fixture_data/bls12-381-sha-256/proof/proof001.json
tplooker: Good point have captured notes of this in #225 and #7

tooling/fixtures/fixture_data/bls12-381-sha-256/signature/signature001.json
tplooker: Have captured a note to follow up on this in #7

tplooker:
I've attempted to capture all the comments from this PR in #225 and #7. Because we have three implementations verifying these fixtures I'd like to merge at this point and publish a new draft revision ahead of the IETF meeting.

tplooker:
Multiple approvals, reviewed on last WG call, residual gaps captured in issues, merging

christianpaquin:
Great. I updated my lib to fetch fixtures from the main branch.

PR #222, opened 2022-10-14
Replace 0 (zero) with O in suite id

tmarkovski:
I'm assuming the 0 was not intended to be there. Can easily cause headaches when copy/pasting from the spec page


andrewwhitehead (approved):

BasileiosKal (approved):

tplooker (approved):

tplooker:
Discussed on WG call 17th of October, multiple approvals, merging

PR #226, opened 2022-10-21
[224] Update D def to simplify calc.

### LINKED ISSUE #224: Enhancement proposal: change D sign in the defenition
The proof gen procedure [defines](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#name-proofgen) `D = B * r1 + Q_1 * r2`.

Everywhere below it's used with `-`.

Can we change the sign of the definition and declare `D = -(B * r1 + Q_1 * r2)` ?

Then the step 17th will be 
`17. C2 = D * r3~ + Q_1 * s~ + H_j1 * m~_j1 + ... + H_jU * m~_jU`

Then proof verification steps 10, 12 will be:
`10. C1 = (Abar + D) * c + A' * e^ + Q_1 * r2^`
`12. C2 = T * c + D * r3^ + Q_1 * s^ + H_j1 * m^_j1 + ... + H_jU * m^_jU`

UPD:
as @tplooker mentioned in the related MR #226, this is a breaking changes, as D is a part of serialized proof.

jovfer:
Fix #224

### DIFF
--- a/draft-irtf-cfrg-bbs-signatures.md
+++ b/draft-irtf-cfrg-bbs-signatures.md
@@ -526,10 +526,10 @@ Procedure:
 11. r3 = r1 ^ -1 mod r
 12. A' = A * r1
 13. Abar = A' * (-e) + B * r1
-14. D = B * r1 + Q_1 * r2
+14. D = -(B * r1 + Q_1 * r2)
 15. s' = r2 * r3 + s mod r
 16. C1 = A' * e~ + Q_1 * r2~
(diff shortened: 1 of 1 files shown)


tplooker:
Thanks @jovfer, considering that this is a breaking change to the proof representation and it only provides a minor improvement around proof computation we will keep this PR open until we consider other breaking changes that it could be adopted with.

tplooker:
Discussed on the 12th of December WG call, given the improvement in readability for this proposal is minor and the performance difference is likely to be negligible we are not going to proceed with this change for now, closing the PR on that basis, please re-open if you have any further questions.

Issue #227, opened 2022-10-31
Problem in ProofVerify


rolfhaenni:
In the current version of ProofVerify, the following problem may arise:

If the given parameter L and the number of disclosed messages R=length(disclosed_indexes) are used for computing U=L-R, then this U may be different from the length the vector (m^_j1,...,m^_jU) included in the proof. Without taking further actions, computing the sum-of-products H_j1 * m^_j1 + ... + H_jU * m^_jU  in Line 12 may then fail in an actual implementation, for example with an IndexOutOfBoundException.

What is missing is an additional precondition R + length(m^_j1,...,m^_jU) = L.

Even a better solution is removing parameter L from the algorithm's parameter list and computing R=length(disclosed_indexes), U=length(m^_j1,...,m^_jU), and L=R+U. This avoids the additional precondition and simplifies the interface.

tmarkovski:
Good point, the L argument seems redundant, since it can be calculated from the submitted proof and messages input. However, this slightly complicates the logic for extracting the commitments from the proof, since their total number is unknown. Implementers would have to ensure that the total number of commitments in m^_jX is the left-over octet length / scalar length. Not a huge deal, but it may complicate the operation definition a little.

This means the integrity of the API method definition is dependent on the encoded data provided inside one of it's arguments. I do tend to favor api definition simplicity, but I'm unsure about this specific case.
Adding the additional precondition seems the most straightforward fix.

tplooker:
Discussed on WG call 21st of November, the proposal seems like an improvement to the specification. I'm going to mark this issue as ready for PR.

BasileiosKal:
Taking the time to look into this proposal a bit more closely, i do like it. I'm not sure how the current approach has an issue problem. If an error is returned (like an IndexOutOfBoundException) it's not a bad thing necessarily, since in this case the proof is malformed.

That said i do like removing `L` from the input of `ProofVerify` as well as separating the deserialization in their own steps. 

Just a minor editorial detail, it seems to me that the `Precomputations` steps loose their purpose a bit (operations to be cached). Maybe we can move more staff to `Deserialization`. Maybe something like:


```
result = ProofVerify(PK, proof, header, ph, disclosed_messages, disclosed_indexes)

Outputs:
- result, either VALID or INVALID. 

Deserialization:
1.  proof_result = octets_to_proof(proof)
2.  (A', Abar, D, c, e^, r2^, r3^, s^, commitments) = proof_result 
[...]

PR #230, opened 2022-12-12
Encode for hash update

### LINKED ISSUE #225: Encode for hash revision
As members of the WG have noted (thanks @christianpaquin, @tmarkovski!!), the current [EncodeForHash](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#name-encodeforhash) section is confusing and prone to error. Especially the part about encoding octet strings.

A possible solution is to remove entirely the part about appending octet strings with their length. The only place this is actually used is for the `msg` in `MapMsgToScalarAsHash`, for the `header` in the domain hash and the `presentationHeader` in the challenge hash. I believe we can handle those cases in their respected functions. Octet string will then be encoded directly.

- **Pros**: will make `EncodeForHash` much simpler + no need for the Ciphersuite ID and Public Key exceptions.
- **Cons**: Will add 1-2 additional step to each of the core operations.

IMO the tradeoff is worth it, but keen hear other approaches as well.

BasileiosKal:
Fixes #225

1. Updated `encode_for_hash` per @andrewwhitehead's proposal from [here](https://github.com/decentralized-identity/bbs-signature/issues/225#issuecomment-1328315289) (including renaming the operation to serialize and definING `SignatureToOctets` and `ProofToOctets` based on `serialize`).
2. Added 2 new operations, `get_domain` and `get_challenge`. Their purpose is to simplify the core operations, since with the revision on encode_for_hash they would have become more complex (note [proofGen](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#name-proofgen) will be 32 steps instead of 24 without those 2 operations).

**NOTE**: I would like to introduce a breaking change and put the PK at the end of the serialized array in `get_domain` (step 7), to be consistent with the “octet strings go at the end rule”. Not worth it now, but maybe will bundle it together with some other breaking changes at some point.

Not sure if the result is simpler overall but is more rigorous and much better defined IMO.


draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -265,6 +268,23 @@ Throughout the operations of this signature scheme, each message that is signed
...
 Aside from the message generators, the scheme uses two additional generators: `Q_1` and `Q_2`. The first (`Q_1`), is used for the blinding value (`s`) of the signature. The second generator (`Q_2`), is used to sign the signature's domain, which binds both the signature and generated proofs to a specific context and cryptographically protects any potential application-specific information (for example, messages that must always be disclosed etc.).
 
+### Serializing to octet strings
+
+When serializing different types of elements (i.e., Points, Scalars etc.) to get their octet representation, each element will be serialized with a specific operation, depending on its type. More concretely,
+
+- Points in `G*` will be serialized using the `point_to_octets_g*` implementation for a particular ciphersuite.
+- Non-negative integers will be serialized using `I2OSP` with an output length of 8 bytes.
+- Scalars will be serialized using `I2OSP` with a constant output length defined by a particular ciphersuite.
+- ASCII strings will be transformed into octet strings using UTF-8 encoding.
```
andrewwhitehead, line 278:
```
- ASCII strings will be transformed into octet strings using UTF-8 encoding.
```

This reads a little strangely as there is no transformation involved, ASCII strings are UTF-8 strings.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -265,6 +268,23 @@ Throughout the operations of this signature scheme, each message that is signed
...
+
+Variable length octet strings, will be prepended with a value representing the length of their binary representation in the form of the number of bytes. This length must be encoded to octets using I2OSP with output length of 8 bytes. The combined value (encoded value + length prefix) binary representation is then encoded as a single octet string. For example, the string `0x14d` will be encoded as `0x0000000000000002014d`. If the length of the octet string is larger than 2^64 - 1, the octet string must be rejected. Similarly, ASCII strings, after encoded to octets (using utf8), will also be appended with the length of their octet-string representation.
```
andrewwhitehead, line 280:
```suggestion
Variable-length octet strings will be prepended with an integer value representing the number of bytes in the string. This length is encoded to octets using `I2OSP` with an output length of 8 bytes. For example, the octet string `0x14d` is encoded as `0x0000000000000002014d`. If the length of the octet string exceeds `2^64 - 1`, the octet string must be rejected. ASCII strings are also encoded using this process by first converting to an octet string.
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -265,6 +268,23 @@ Throughout the operations of this signature scheme, each message that is signed
...
+
+Constant length octet strings (e.g., the ciphersuite id octet representation), will be encoded directly.
```
andrewwhitehead, line 282:
```suggestion
Constant-length octet strings, including domain separation tags and the ciphersuite ID, will be encoded directly without prefixing with the length.
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -265,6 +268,23 @@ Throughout the operations of this signature scheme, each message that is signed
...
+
+Optional input/parameters to operations that feature in a call to hash\_to\_scalar, that are not supplied to the operation should default to an empty octet string. For example, if X is an optional input/parameter that is not supplied, whilst A and B are required, then the procedural step of `hash(A || X || B)` MUST be evaluated to `hash(A || "" || B)`.
```
andrewwhitehead, line 284: I think we should just be explicit about what the defaults are for each variable at the point of use

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -265,6 +268,23 @@ Throughout the operations of this signature scheme, each message that is signed
 
 Aside from the message generators, the scheme uses two additional generators: `Q_1` and `Q_2`. The first (`Q_1`), is used for the blinding value (`s`) of the signature. The second generator (`Q_2`), is used to sign the signature's domain, which binds both the signature and generated proofs to a specific context and cryptographically protects any potential application-specific information (for example, messages that must always be disclosed etc.).
 
+### Serializing to octet strings
+
+When serializing different types of elements (i.e., Points, Scalars etc.) to get their octet representation, each element will be serialized with a specific operation, depending on its type. More concretely,
```
andrewwhitehead, line 273:
```suggestion
When serializing one or more values to produce an octet string, each element will be encoded using a specific operation determined by its type. More concretely,
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -779,8 +787,158 @@ Procedure:
...
 ```
 
+## Domain Calculation
+
+This operation calculates the domain value, a scalar encoding public parameter information. Optionally, the domain can also include an additional octet string, called the header, meant to encode any additional information the signer may want to "bind" to the signature (like an expiration date, the targeted audience identifier etc.). The domain value will be signed using a specific generator point (`Q_2`) (see [Sign](#sign)) to protect the integrity of those public parameters. The header, if existing, must be known by all parties (i.e., the signer, the prover and verifier), for both the signature and proof to be validated correctly. This means, that the header can be used to enforce certain values (chosen by the signer) to always be disclosed by the prover.
+
+This operation makes use of the `serialize` function, defined in [Section 4.6.1](#serialize).
+
+```
+domain = get_domain(Q_1, Q_2, H_Points, PK, header)
```
andrewwhitehead, line 797: Calls to this method have PK as the first argument.

andrewwhitehead: Maybe `encode_domain` (and `encode_challenge`) would be more descriptive?

andrewwhitehead: Or `derive_`, `calculate_`, `hash_`, etc...

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -779,8 +787,158 @@ Procedure:
 11. return (scalar_1, ..., scalar_count)
 ```
 
+## Domain Calculation
+
+This operation calculates the domain value, a scalar encoding public parameter information. Optionally, the domain can also include an additional octet string, called the header, meant to encode any additional information the signer may want to "bind" to the signature (like an expiration date, the targeted audience identifier etc.). The domain value will be signed using a specific generator point (`Q_2`) (see [Sign](#sign)) to protect the integrity of those public parameters. The header, if existing, must be known by all parties (i.e., the signer, the prover and verifier), for both the signature and proof to be validated correctly. This means, that the header can be used to enforce certain values (chosen by the signer) to always be disclosed by the prover.
```
andrewwhitehead, line 792:
```suggestion
This operation calculates the domain value, a scalar representing the distillation of all essential contextual information for a signature. The same domain value must be calculated by all parties (the signer, the prover, and the verifier) for both the signature and proofs to be validated correctly.

The input to the domain value includes an octet string called the header, chosen by the signer and meant to encode any information that is required to be revealed by the prover (such as an expiration date, or an identifier for the target audience). This is in contrast to the signed message values, which may be withheld during a proof.

When a signature is calculated, the domain value is combined with a specific generator point (`Q_2`, see [Sign](#sign)) to protect the integrity of the public parameters and the header.
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -779,8 +787,158 @@ Procedure:
...
+2. if length(header) > 2^64 - 1 or L > 2^64 - 1, return INVALID
+3. (H_1, ..., H_L) = H_Points
+4. dom_array = (Q_1, Q_2, L, H_1, ..., H_L, ciphersuite_id)
+5. dom_octs = serialize(dom_array)
+6. if dom_octs is INVALID, return INVALID
+7. dom_input = PK || dom_octs || I2OSP(length(header), 8) || header
+8. return hash_to_scalar(dom_input, 1)
+```
+
+**Note**: If the header is not supplied in `get_domain`, it defaults to the empty octet string (""). This means that in the concatenation step of the above procedure (step 7), 8 bytes representing a length of 0 (i.e., `0x0000000000000000`), will still need to be appended at the end, even though the a header value is not provided.
```
andrewwhitehead, line 829:
```suggestion
**Note**: If the header is not supplied in `get_domain`, it defaults to the empty octet string ("").
```

I think it's clear in the algorithm that the length is always included

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -779,8 +787,158 @@ Procedure:
...
+Outputs:
+
+- domain, a scalar value or INVALID.
+
+Procedure:
+
+1. L = length(H_Points)
+2. if length(header) > 2^64 - 1 or L > 2^64 - 1, return INVALID
+3. (H_1, ..., H_L) = H_Points
+4. dom_array = (Q_1, Q_2, L, H_1, ..., H_L, ciphersuite_id)
```
andrewwhitehead, line 822: Shouldn't `ciphersuite_id` be appended later and not passed to `serialize`?

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -779,8 +787,158 @@ Procedure:
...
+                                                  msg_array, domain, ph)
+
+Inputs:
+
+- (A', Abar, D, C1, C2) (REQUIRED), points of G1, as calculated in
+                                    ProofGen
+- i_array (REQUIRED), array of non-negative integers (the indexes of
+                      the disclosed messages).
+- msg_array (REQUIRED), array of scalars (the disclosed messages).
+- domain (REQUIRED), octet string.
```
andrewwhitehead, line 848:
```suggestion
- domain (REQUIRED), a scalar.
```

draft-irtf-cfrg-bbs-signatures.md
andrewwhitehead: Ah, I see there is a rule for ASCII strings.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -810,32 +968,38 @@ Procedure:
...
+- proof_octets, octet string or INVALID.
 
 Procedure:
 
-1. (A, e, s) = signature
-2. A_octets = point_to_octets_g1(A)
-3. e_octets = I2OSP(e, octet_scalar_length)
-4. s_octets = I2OSP(s, octet_scalar_length)
-5. return (A_octets || e_octets || s_octets)
+1. (A', Abar, D, c, e^, r2^, r3^, s^, (m^_1, ..., m^_U)) = proof
```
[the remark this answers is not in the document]
BasileiosKal: Not a bad idea to add some redundancy in the proof value. Would also make the suggestion in #227 easier to implement.

That said, since `ProofVerify` is not a terribly inefficient operation, not sure how much will provide beyond some useful error information? Also, is truncating the proof something that will (somewhat) "commonly" happen (i assume not as much in the application layer)??

In any case it seems that it would be better addressed in a separate PR, so will open an issue to track the conversation and will resolve this comment if that is ok? Please feel free to un-resolve!

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: I do agree with this. The current rule followed in the doc, is that ASCII string refers to text and ASCII code refers to bytes (note the ciphersuite id and DSTs encodings).

We could make this clearer by updating ASCII string -> Text or ASCII text or String??

Since this affects the whole doc, IMO its better to open a issue for tracking and address it in a separate PR?

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: That's true, though IMO it would be an easy mistake to make.

its clear only if you notice the `If not supplied, it must default to the empty octet string ("")` part. If someone misses that and only notice the capitalized `(OPTIONAL)` could maybe make this mistake??

BasileiosKal: maybe we could update the text to say `If not supplied, it must default to the empty octet string ("") **that has 0 length**`

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Done! (used `calculate_`).

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Agreed. Removed the paragraph on optional inputs. For constant length octet strings will revisit after #232 is resolved.

draft-irtf-cfrg-bbs-signatures.md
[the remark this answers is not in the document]
BasileiosKal: opened #233 to track the issue!

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Opened #234 to track this issue!

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Yea. I was trying to follow the "encode everything that are not already octet strings" rule. I agree that this may not be optimal though. Opened #234 to discuss further.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -152,7 +155,7 @@ presentation\_header (ph)
 : A payload generated and bound to the context of a specific spk.
 
 nizk
-: A non-interactive zero-knowledge proof from fiat-shamir heuristic.
+: A non-interactive zero-knowledge proof from Fiat-Shamir heuristic.
```
tplooker, line 158:
```suggestion
: A non-interactive zero-knowledge proof from the Fiat-Shamir heuristic.
```

tplooker (approved):

andrewwhitehead (approved):

BasileiosKal:
Discussed on the WG call 9th of January. Multiple approvals. Merging. Thank you everyone!

PR #236, opened 2023-01-19
Use CSPRNGs instead of PRFs

### LINKED ISSUE #229: PRF vs RNG/RBG
IMO we are little vague regarding the PRFs in the spec.

For example, in RFCs like [IKEv2](https://www.rfc-editor.org/rfc/rfc7296) and [RFC4868](https://www.rfc-editor.org/rfc/rfc4868.html) (which we link to as a reference), PRFs are keyed, accept additional input and return constant size output. However, we do call them just on the required output len.

You can build an extendable output PRF, (or PRF* as is called in [[H2010]](https://link.springer.com/content/pdf/10.1007/978-3-642-14623-7_34.pdf)), using a PRF in counter/feedback mode etc., and get something usable by our operations, however we don’t mention that currently.

I think what we are using is something closer to an RBG or a RNG (or a CSPRBG/CSPRNG more specifically) than a PRF.

Is there a reason to use a PRF?? The differences are subtle (and of-course you can build RBGs/RNGs on top of PRFs and vice-versa). It seems more confusing and harder to work with to me though. For example, with PR #228 we recommend the use of a seeded ChaCha20 DRNG to avoid some entropy attacks. However, the result is not "strictly speaking" a PRF.

Other specs like [FROST](https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-11.html) or [OPAQUE](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-09.html) also use a CSPRNG with both pointing to [RFC4086](https://www.rfc-editor.org/rfc/rfc4086) for requirements.

BasileiosKal:
Closes #229


christianpaquin (approved):

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -254,7 +254,7 @@ The schemes operations defined in this section depend on the following parameter
 
 * A hash-to-curve suite as defined in [@!I-D.irtf-cfrg-hash-to-curve], using the aforementioned pairing-friendly curve. This defines the hash\_to\_curve and expand\_message operations, used by this document.
 
-* PRF(n): a pseudo-random function similar to [@!RFC4868]. Returns n pseudo randomly generated bytes.
+* get\_random(n): returns a random octet string with a length of n bytes, sampled uniformly at random using a cryptographically secure pseudo-random number generator (CSPRNG) or a pseudo random function. See [@!RFC4086] for recommendations and requirements on the generation of random numbers
```
andrewwhitehead, line 257:
```suggestion
* get\_random(n): returns a random octet string with a length of n bytes, sampled uniformly at random using a cryptographically secure pseudo-random number generator (CSPRNG) or a pseudo random function. See [@!RFC4086] for recommendations and requirements on the generation of random numbers.
```

Wind4Greg (approved):

tplooker (approved):

tplooker:
Multiple reviews and approvals, merging.

PR #237, opened 2023-01-20
Keygen tool and fixtures update

BasileiosKal:
As noted from @Wind4Greg who i thank a lot, the keyPair fixtures did not validate using the keyGen operation described in the spec. The reason was a mistake i made in the implementation that was used to create those fixtures. 

This PR:

- Updates the KeyPair fixture (also adds the `IKM` and `key_info` inputs).
- Updates all the fixtures to use the correct KeyPair.
- Adds a reference tool that can be used to generate and save BBS KeyPairs.

Will update after #186 is resolved. Thanks again @Wind4Greg! And sorry everyone for the inconvenience of having to update!


Wind4Greg (approved):
Verified updated test vectors for key generation against draft procedures in JavaScript and Python.

tplooker (approved):

tplooker:
Multiple reviews and approvals, merging as it is an important update to fixtures that is aligned to current spec text

PR #238, opened 2023-01-20
Removing L from ProofVerify input

### LINKED ISSUE #227: Problem in ProofVerify

BasileiosKal:
Closes #227

Taking a stab at the approach proposed by @rolfhaenni in #227. This PR:
1.	Removes L from the inputs of ProofVerify and calculates it as L=R+U
2.	Adds a `Deserialization` group of steps in the core operations (also used for de-structuring).

Also removed the `Precomputation` steps in favor of the `Deserialization` ones (IMO what was a `Precomputation` step was a bit loosely defined).

Since I removed the text saying that the `Precomputation` steps can be cached, I added some text detailing how the output of `create_generaotrs` can be cached + some requirements (although we may want to extent upon them in the future).

We could potentially also remove the`Definitions` from each operation, since those values are explicitly defined in the procedures.

Thanks again @rolfhaenni 🙌


draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -598,49 +598,55 @@ Parameters:
...
+6.  U  = length(commitments)
+7.  R = length(disclosed_indexes)
+8.  L = R + U
+9.  (i1, ..., iR) = disclosed_indexes
+10. (j1, ..., jU) = range(1, L) \ disclosed_indexes
+11. (msg_i1, ..., msg_iR) = disclosed_messages
+12. (m^_j1, ...., m^_jU) = commitments
 
-Preconditions:
+Preconditions:s
```
christianpaquin, line 626: Typo?

Wind4Greg (approved):

draft-irtf-cfrg-bbs-signatures.md
tplooker:
```suggestion
Preconditions:
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -649,7 +655,13 @@ Procedure:
...
 
-As an optimization, implementations MAY cache the result of `create_generators` for a specific `generator_seed` (determined by the ciphersuite) and `count`. The values `n` and `v` MAY also be cached in order to efficiently extend a existing list of generator points.
+As an optimization, implementations MAY cache the result of `create_generators` for a specific `generator_seed` (determined by the ciphersuite) and `count` (which can be arbitrarily large, depending on the application). Then, during the execution of one of the [Core Operations](#core-operations), if `K` generators are needed with `K <= count`, the application can use the `K` first of the cached generators (in place of the direct call to `create_generators(K)`).
+
+**NOTE**: If cached, the order with which the generator points will be retrieved from the cache MUST be the same as the order they where originally returned by the `create_generators` operation.
```
tplooker, line 660:
```suggestion
**NOTE**: If the generator points are retrieved from cache, the order in which they are retrieved MUST be the same as the order they were originally returned by the `create_generators` operation.
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -649,7 +655,13 @@ Procedure:
...
+
+For example, an application can save 100 generator points `H_1, H_2, ..., H_100` returned from `create_generators(100)`. Then if one of the core operations needs 30 of them, the application instead of calling `create_generators` again, can just retrieve the 30 first generators `H_1, H_2, ..., H_30` from the cache instead, in the same order they where originally created (starting from the first one).
+
+The values `n` and `v` MAY also be cached in order to efficiently extend a existing list of cached generator points.
```
tplooker, line 664:
```suggestion
The values `n` and `v` MAY also be cached in order to efficiently extend an existing list of cached generator points.
```

tplooker (approved):

tplooker:
Multiple approvals, open 3 weeks, discussed on WG calls, merging

PR #239, opened 2023-01-23
Deterministic proof fixtures

### LINKED ISSUE #73: Deterministic SPK
Although deterministic proof generation should not be used in practice it may be useful to support it, so we can more easily create and update test vectors. This MUST NEVER be used in practice. The main idea is to incorporate a secret key managed by the holder of the bbs signature into spkGen that is going to be used for the generation of the random elements. This secret MUST be unique for every proof generation procedure. One possible method then, is to use the same algorithm as in deterministic signing, i.e.,
```
H = XOF(Holders_SK || msg_i1 ||…||msg_iR || pm)
r1~ = OS2IP(H.read(64)) mod q
r2~ = OS2IP(H.read(64)) mod q
              .
              .
              .
```
In spkGen, maybe we can also use [RFC 8937](https://www.rfc-editor.org/rfc/rfc8937.html), which gives good security properties and is easy to use, since you don't need direct access to the secret key itself. Each random element (in the non-deterministic case) could be
```
random_element_i = OS2IP(HKDF-Expand(HKDF-Extract(HASH(Sig(Holders_SK, DST)), PRF(L)), i, 64)) mod q
```
To get a deterministic spk, we could just replace the PRF with an XOF or hash. That said, this does "bound" the holder to use a secret, even in the non-deterministic case.

BasileiosKal:
closes #73, #7 

Describing a mocked seeded rng, using `expand_message`, which will be used to describe proof test vectors.


draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1290,6 +1320,61 @@ The following section details a basic set of test vectors that can be used to co
...
+mocked_calculate_random_scalars(count) :=
+                             seeded_random_scalars(count, SEED)
+```
+
+in place of `calculate_random_scalars` during the operation's procedure.
+
+**Note** For the [BLS12-381-SHA-256](#bls12-381-sha-256) ciphersuite if more than 170 mocked random scalars are required, the operation will return INVALID. Similarly, for the [BLS12-381-SHAKE-256](#bls12-381-shake-256) ciphersuite, if more than 1365 mocked random scalars are required, the operation will return INVALID. For the purpose of describing [ProofGen](#proofgen) test vectors, those limits are inconsequential.
+
+```
+seeded_scalars = seeded_random_scalars(count, SEED)
```
andrewwhitehead, line 1339:
```suggestion
seeded_scalars = seeded_random_scalars(seed, count)
```

Seems a bit more consistent with the rest of the draft?

Wind4Greg (approved):

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Fixed. Thanks!

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1456,6 +1542,56 @@ And the messages defined in (#messages) (**Note** the ordering of the messages M
 {{ $signatureFixtures.bls12-381-shake-256.signature004.signature }}
 ```
 
+### Proof fixtures
+
+For the generation of the following fixtures the `mocked_calculate_random_scalars` defined in [Mocked Random Scalars](#mocked-random-scalars) is used, in place of the `calculate_random_scalars` operation, with the following seed value (hex encoding of `utf8("<30 first digits of pi>")`)
```
tplooker, line 1547:
```suggestion
For the generation of the following fixtures the `mocked_calculate_random_scalars` defined in [Mocked Random Scalars](#mocked-random-scalars) is used, in place of the `calculate_random_scalars` operation, with the following seed value (hex encoding of `utf8("<first 30 digits of pi>")`)
```

BasileiosKal:
Discussed on WG call on the 14th of Feb. Will await verification of fixtures from implementations.

tplooker (approved):

tplooker:
Multiple approvals, discussed on multiple WG calls, fixtures submitted have now been verified by @Wind4Greg's implementation 🚀

PR #243, opened 2023-02-06
Reduce hash_to_scalar to a single output

### LINKED ISSUE #242: hash_to_scalar is overly flexible
Because it is no longer used to generate multiple outputs except for a pair of `e` and `s`, we can simplify `hash_to_scalar` to output only a single point instead of a list. In `Sign` we might use the following procedure:

```
3. e_s_for_hash = serialize((SK, domain, msg_1, ..., msg_L))
4. if e_s_for_hash is INVALID, return INVALID
5. e = hash_to_scalar(e_s_for_hash || I2OSP(0, 1))
6. s = hash_to_scalar(e_s_for_hash || I2OSP(1, 1))
```

andrewwhitehead:
Resolves #242 

Requires some updates to test vectors.

### DIFF
--- a/draft-irtf-cfrg-bbs-signatures.md
+++ b/draft-irtf-cfrg-bbs-signatures.md
@@ -424,7 +424,7 @@ Procedure:
 13. return signature_to_octets(A, e, s)
 ```
 
-**Note** When computing step 9 of the above procedure there is an extremely small probability (around `2^(-r)`) that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evaluate to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 9, and or `A != Identity_G1` after step 9 to prevent the production of invalid signatures.
+**Note** When computing step 12 of the above procedure there is an extremely small probability (around `2^(-r)`) that the condition `(SK + e) = 0 mod r` will be met. How implementations evaluate the inverse of the scalar value `0` may vary, with some returning an error and others returning `0` as a result. If the returned value from the inverse operation `1/(SK + e)` does evaluate to `0` the value of `A` will equal `Identity_G1` thus an invalid signature. Implementations MAY elect to check `(SK + e) = 0 mod r` prior to step 9, and or `A != Identity_G1` after step 9 to prevent the production of invalid signatures.
 
 ### Verify
 
--- a/draft-irtf-cfrg-bbs-signatures.md
+++ b/draft-irtf-cfrg-bbs-signatures.md
@@ -415,10 +415,10 @@ Procedure:
 4.  e_s_octs = serialize((SK, domain, msg_1, ..., msg_L))
 5.  if e_s_octs is INVALID, return INVALID
 6.  e_s_expand = expand_message(e_s_octs, expand_dst, expand_len * 2)
-7.  e = hash_to_scalar(e_s_expand[0..(expand_len - 1)])
-8.  if e is INVALID, return INVALID
+7.  if es_expand is INVALID, return INVALID
+8.  e = hash_to_scalar(e_s_expand[0..(expand_len - 1)])
 9.  s = hash_to_scalar(e_s_expand[expand_len..(expand_len * 2 - 1)])
-10. if s is INVALID, return INVALID
+10. if e or s is INVALID, return INVALID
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -403,15 +403,16 @@ Precomputations:
...
-7. B = P1 + Q_1 * s + Q_2 * domain + H_1 * msg_1 + ... + H_L * msg_L
-8. A = B * (1 / (SK + e))
-9. return signature_to_octets(A, e, s)
+1.  domain = calculate_domain(PK, Q_1, Q_2, (H_1, ..., H_L), header)
+2.  if domain is INVALID, return INVALID
+3.  e_s_octs = serialize((SK, domain, msg_1, ..., msg_L))
+4.  if e_s_octs is INVALID, return INVALID
+5.  e = hash_to_scalar(e_s_octs || I2OSP(0, 1))
+6.  s = hash_to_scalar(e_s_octs || I2OSP(1, 1))
+7.  if e or s is INVALID, return INVALID
```
BasileiosKal, line 412: In this, the inputs for `e` and `s` differ in one bit only. This will highly correlate the 2 values. Maybe we can do something like,

```
5.  e = hash_to_scalar(e_s_octs || I2OSP(0, 1))
6.  if e is INVALID, return INVALID
7.  s = hash_to_scalar(I2OSP(e) || I2OSP(1, 1))
8.  if s is INVALID, return INVALID
```
This has another small issue, namely that the 2 inputs do not have the same length, which will make it hard to guarantee that the Hamiltonian distance between them is infarct large.

We could argue that since the SK is unknown, in the RO model the 2 inputs will be at least 32 random bytes apart. However ideally we would do something like:
```
5.  v = hash_to_scalar(e_s_octs || I2OSP(0, 1))
6.  if v is INVALID, return INVALID
7.  e = hash_to_scalar(I2OSP(v) || I2OSP(1, 1))
8.  if e is INVALID, return INVALID
9.  s = hash_to_scalar(I2OSP(e) || I2OSP(2, 1))
10. if s is INVALID, return INVALID
```
Here we don't require `hash_to_scalar` to be an RO, just the computational assumptions for the underlying hash will do. But at this point we rather complicate things IMO.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -405,13 +412,16 @@ Procedure:
...
-4.  e_s_for_hash = serialize((SK, domain, msg_1, ..., msg_L))
-5.  if e_s_for_hash is INVALID, return INVALID
-6.  (e, s) = hash_to_scalar(e_s_for_hash, 2)
-7.  if e or s is INVALID, return INVALID
-8.  B = P1 + Q_1 * s + Q_2 * domain + H_1 * msg_1 + ... + H_L * msg_L
-9.  A = B * (1 / (SK + e))
-10. return signature_to_octets(A, e, s)
+4.  e_s_octs = serialize((SK, domain, msg_1, ..., msg_L))
+5.  if e_s_octs is INVALID, return INVALID
+6.  e_s_expand = expand_message(e_s_octs, expand_dst, expand_len * 2)
```
BasileiosKal, line 417: `expand_message` could abort as well. Should we add a check that `e_s_expand` is not INVALID (and maybe a note that if `expand_message` aborts we assume it returns INVALID??)

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -405,13 +412,16 @@ Procedure:
...
+7.  e = hash_to_scalar(e_s_expand[0..(expand_len - 1)])
+8.  if e is INVALID, return INVALID
+9.  s = hash_to_scalar(e_s_expand[expand_len..(expand_len * 2 - 1)])
+10. if s is INVALID, return INVALID
```
BasileiosKal, line 421:
```suggestion
7.  e = hash_to_scalar(e_s_expand[0..(expand_len - 1)])
9.  s = hash_to_scalar(e_s_expand[expand_len..(expand_len * 2 - 1)])
8.  if e or s is INVALID, return INVALID
```
Just a very small suggestion for better readability??

draft-irtf-cfrg-bbs-signatures.md
andrewwhitehead: I had thought we treated it as infallible, but apparently not, so updated now.

BasileiosKal:
Some small nits. LGTM otherwise. It complicates `Sign` a bit more than i would like TBH, but given that we will probably remove `s` eventually, i think this is the better approach (removing `s` will significantly simplify `Sign` again).

BasileiosKal (approved):

tplooker (approved):

tplooker:
Multiple reviews and approvals, discussed on prior WG calls, merging.

PR #251, opened 2023-03-10
Fixed type of serialize's octets_result

christianpaquin:
Fixed type of output value `octets_result`, and added the article "an" before some "octet string" for consistency.


tplooker (approved):

tplooker:
Thanks @christianpaquin I have addressed in this commit https://github.com/decentralized-identity/bbs-signature/commit/ef22de1afdb3cda4b7b8d16cef0fc072bd7bcdbc as the PR has conflicts and we need to submit draft-02 before the deadline.

PR #255, opened 2023-04-07
Added text on header, ph usage.

Wind4Greg:
Per the last meeting I added some small clarifying text on the `header` and `ph` fields.

Cheers
Greg

### DIFF
--- a/draft-irtf-cfrg-bbs-signatures.md
+++ b/draft-irtf-cfrg-bbs-signatures.md
@@ -377,7 +377,8 @@ Inputs:
                  operation.
 - PK (REQUIRED), an octet string of the form outputted by the SkToPk
                  operation provided the above SK as input.
-- header (OPTIONAL), an octet string containing context and application
+- header (OPTIONAL), an octet string containing protected associated data 
+                     such as context and application
                      specific information. If not supplied, it defaults
                      to an empty string.
 - messages (OPTIONAL), a vector of scalars. If not supplied, it defaults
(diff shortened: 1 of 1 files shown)


tplooker:
Discussed on the WG call, decided that the best way to handle this caveat is to add a privacy consideration section to the spec that talks to this specific issue around header usage.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -377,7 +377,8 @@ Inputs:
                  operation.
 - PK (REQUIRED), an octet string of the form outputted by the SkToPk
                  operation provided the above SK as input.
-- header (OPTIONAL), an octet string containing context and application
+- header (OPTIONAL), an octet string containing protected associated data 
+                     such as context and application
```
BasileiosKal, line 381: IMO, those descriptions are not really suited for the input section of the operations, where we should try to keep things as simple as possible. I think there are 2 alternatives:
1. Add the `header` and `presentation_header` terms in the [Terminology section](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#name-terminology), followed by the above description.
2.  Add a new subsection of the [Considerations section](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#name-considerations), to describe their functionality in more detail.

(I would go with option 2)

This is in addition to the privacy section around the header/ph use, mentioned [here](https://github.com/decentralized-identity/bbs-signature/pull/255#issuecomment-1502141929)

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -502,7 +503,8 @@ Inputs:
 - header (OPTIONAL), an octet string containing context and application
                      specific information. If not supplied, it defaults
                      to an empty string.
-- ph (OPTIONAL), an octet string containing the presentation header. If not
+- ph (OPTIONAL), an octet string containing the presentation header which 
+                 is protected associated data. If not
```
BasileiosKal, line 507: Same as above^

BasileiosKal (commented):
Thanks Greg! The descriptions look great. Just a comment r.e., their placement in the document.

draft-irtf-cfrg-bbs-signatures.md
Wind4Greg: I like your options 2. Cheers Greg

Issue #262, opened 2023-04-30
Bound BBS signatures


BasileiosKal:
Hey all!

Posting a first draft of bound BBS signatures, using BLS key pairs: https://github.com/BasileiosKal/bbs-bound-signatures/blob/main/draft-bound-bbs-signatures.md

(there is also an html version in the [repo](https://github.com/BasileiosKal/bbs-bound-signatures))

OR13:
Can you make the html link show up in the github page fro the repo?

tplooker:
Discussed on WG call 1st of May, In order to better support this draft we have identified a couple of places that the core draft could be better, @BasileiosKal will raise some issues to capture these concepts.

Wind4Greg:
Hi Vasilis, I did an initial read. I think the curve BLS12-381 curve library I use has a pretty good BLS signature implementation. Let us know if/when you've got some test vectors and I can try a JavaScript implementation. Cheers Greg

BasileiosKal:
@OR13 made the link available. You can find it [here](https://basileioskal.github.io/bbs-bound-signatures/draft-bound-bbs-signatures.html) or from the [README](https://github.com/BasileiosKal/bbs-bound-signatures/blob/main/README.md). Thank you for the recommendation!

BasileiosKal:
@Wind4Greg That's awesome!! We have an implementation of the draft so we will be able to post some test vectors soon. Thank you!

BasileiosKal:
Thanks for the feedback @andrewwhitehead 🙏 

> It seems like the verification is missing a way to pass in the BP_1 generator point to BbsVerify. 

BbsVerify should be using this `create_generators`. Wouldn't this be enough??

> I'm a little unsure about just adding BBS_BOUND to the header instead of defining a new ciphersuite, 

I agree. This was a temporally solution until we define a new ciphersuite. The structure of a bound ciphersuite will depend on the flexibility of the core draft, so waiting for that discussion there to move on a bit first.

> For our applications I think I would also prefer a zero-knowledge proof of key possession and blind signing similar to the old implementation, but that would lose the benefit of delegating to the BLS draft.

As an alternative, we could also make one document that encapsulates both generic Schnorr-like commitments and BLS signatures. The flow is the same, so we could define `commitmentGen` in a `commitmentVerify` in a generic way.

andrewwhitehead:
> BbsVerify should be using this create_generators. Wouldn't this be enough??

[...]

PR #264, opened 2023-05-22
Create generators flexibility

### LINKED ISSUE #206: create_generators flexibility

BasileiosKal:
Closes #206

Going with option 2 from the issue, allowing the definition of new `create_generators` and listing the relative requirements. Τhis will also benefit use cases like [Bound signatures](https://github.com/decentralized-identity/bbs-signature/issues/262).

### DIFF
--- a/draft-irtf-cfrg-bbs-signatures.md
+++ b/draft-irtf-cfrg-bbs-signatures.md
@@ -1256,7 +1256,7 @@ The following section defines the format of the unique identifier for the cipher
 
   *  H2C\_SUITE\_ID is the suite ID of the hash-to-curve suite used to define the hash_to_curve function.
 
-  *  CG\_ID is the ID of the create generators used, i.e., `CREATE_GENERATORS_ID` as defined in the [Create Generators](#generator-point-computation) section.
+  *  CG\_ID is the ID of the create generators used, i.e., `CREATE_GENERATORS_ID` as defined in the (#generators-calculation) section.
 
   *  ADD\_INFO is an optional octet string indicating any additional information used to uniquely qualify the ciphersuite. When present this value MUST only contain ASCII encoded characters with codes between 0x21 and 0x7e (inclusive) and MUST end with an underscore (ASCII code: 0x5f), other than the last character the string MUST not contain any other underscores (ASCII code: 0x5f).
 
@@ -1292,7 +1292,7 @@ a function that returns the point P in the subgroup G2 corresponding to the cano
 
 **Generator parameters**:
 
-- create\_generators: the operation with which to create a set of generators. See (#generator-point-computation).
+- create\_generators: the operation with which to create a set of generators. See (#generators-calculation).
 
 ## BLS12-381 Ciphersuites
 
--- a/draft-irtf-cfrg-bbs-signatures.md
+++ b/draft-irtf-cfrg-bbs-signatures.md
@@ -390,6 +390,8 @@ Parameters:
                   specified by the hash_to_curve_suite parameter.
 - octet_scalar_length, non-negative integer. The length of a scalar
                        octet representation, defined by the ciphersuite.
+- create_generators, an operation that returns a number of generator
+                     points, defined by the ciphersuite.
 
 Definitions:
 
@@ -450,6 +452,8 @@ Inputs:
 Parameters:
 
 - P1, fixed point of G1, defined by the ciphersuite.
+- create_generators, an operation that returns a number of generator
+                     points, defined by the ciphersuite.
 
 Definitions:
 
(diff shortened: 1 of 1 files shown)


BasileiosKal:
Discussed on the WG call on the 29th of May, will re-consider the use of `extra_info` and add the PK as a parameter to the create_generators API.

Wind4Greg (approved):

christianpaquin:
I'm not sure I understand the goal of parameterizing the generators creation. Isn't there a value for applications to pre-generate and reuse them across many issuers (which would not be possible of if they depend on an issuer PK).  Is the goal of the `header` to further scope the generators, so an issuer can have multiple distinct generators? Generators can be used to track users, so their uniformity has value.

andrewwhitehead:
The ciphersuite does not have to take the PK and header into account when creating the generators, and the ones defined by the spec will not. It just seemed reasonable to pass those parameters in case the suite wants to define the generators differently (such as how the original implementation did, based on the issuer PK and number of messages). I don't think this increases correlation as the PK and header are known to all verifiers.

andrewwhitehead:
It's the signature header, which is per issuance (and may be empty). It contributes to the domain hash.

christianpaquin:
> It's the signature header, which is per issuance (and may be empty). It contributes to the domain hash.

Ok, so it doesn't add traceability risks then.

BasileiosKal:
I think a malicious issuer could use a unique header per signature to track the user. There are plans for adding privacy considerations to address this. The user (or anyone else) could check the header and decide if it is conformant to those considerations. 

Indented use-cases for the header (e.g., to include algorithm/protocol identifiers etc.,) make it hard to define it as part of the ciphersuite.

BasileiosKal:
Discussed on the WG call on the 19th of June. Will remove the header parameter from the Inputs of `create_generators`.

christianpaquin:
> Discussed on the WG call on the 19th of June. Will remove the header parameter from the Inputs of `create_generators`.

The fixtures from PR #269 currently used the header, right?

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -745,12 +766,29 @@ Procedure:
...
+- It should be infeasible to guess the discrete logarithm of the returned points, for any base, even with knowledge of the public parameters that were used to create those generators (like the `generator_seed` value in [Hash to Generators](#hash-to-generators)). Note that pseudo randomness does not necessarily imply this property. For example, an implementation that repeatably hashes a public seed value to create exponents `r_1, r_2, ..., r_count` (where `r_1 = hash(seed), r_2 = hash(r_1), ...`) and then returns the points `H_1 = P1 * r_1, H_2 = P_1 * r_2, ..., H_count = P_1 * r_count` would be insecure (given knowledge of the seed), but given knowledge of only the points `H_1, ..., H_count`, the sequence would appear random.
+- The returned points must be different from the Identity point of G1 as well as the constant point `P1` defined by the ciphersuite.
+- Must be constant time for a specific `count` value.
+- MUST be deterministic.
+- Must use proper domain separation for both the `create_generator` procedure, as well as all the internally called procedures.
```
andrewwhitehead, line 785:
```suggestion
- Must use proper domain separation for both the `create_generators` procedure, as well as all of the internally-called procedures.
```

andrewwhitehead (approved):

christianpaquin (approved):
LGTM

tplooker (approved):

tplooker:
Discussed on WG call 26th of June, multiple approvals merging.

PR #267, opened 2023-06-12
Error handling

BasileiosKal:
Implicitly handle errors form the utility sub-routines in the core operations + address point **B** from #246

PR #266, opened 2023-06-12
PK format and validation

BasileiosKal:
Addressing point -**A**- from issue #246

PR #268, opened 2023-06-14
Message to Scalar flexibility

### LINKED ISSUE #245: Message to scalar flexibility
Likely, different applications will need different ways to map messages to scalars.

Things to consider:

## How flexible we want to be?

The most flexible option would be to allow the position of a message to affect its encoding i.e., first message using `hash-to-scalar`, second message is zero centered etc. One way to do this would be with some short of a "schema". However an easier way could be to update the `MapMessageToScalar` interface to accept all the messages (instead of having to call that operation for each message separately). Specifically, the generic `MapMessageToScalar` will be defined as,

```
MapMessageToScalar(messages, dst)
```

An application can then define,

```
AppX_MapMessageToScalar(messages, dst)

Deserialization:

1. (msg_1, msg_2, ..., msg_L) = messages

Procedure:
1. scalar_1 = hash_to_scalar(msg_1)
2. scalar_2 = OS2IP(msg_2) mods r
3. ....
```

The `MapMessageToScalarAsHash` operation will then be defined as,

```
MapMessageToScalarAsHash(messages, dst)

Deserialization:

1. (msg_1, msg_2, ..., msg_L) = messages

Precondition

1. If length(msg_i) > 2^64 - 1 for i in (1, ..., L), return INVALID
2. if length(dst) > 255, return INVALID

Procedure:

1. for msg in (1, ..., L)
4.     msg_scalar_i = hash_to_scalar(msg_i, dst)
5.     if msg_scalar_i is INVALID, return INVALID
6. return (msg_scalar_1, ...., msg_scalar_L)
```

This will allow for that level of flexibility without having to define "schemas".

## Who can define a `MapMessageToScalar` operation

Assuming that we would require every `MapMessageToScalar` operation to have a unique ID, there are going to be 2 options here:

### A. Any application can define a new `MapMessageToScalr` operation.
The operations ID would then be added as a parameter to each of the core operations, and then passed to the `domain` value.

### B. Only ciphersuites can define a new `MapMessageToScalar` opeartion.
The operations ID would then just be added to the end of the ciphersuite ID.

IMO, option B is better. If an application "needs" to define a new `MapMessageToScalar` procedure, its likely that they will use it for some cryptographic functionality (like predicate proofs) that would be best to be defined in some document either way (and define its own ciphersuites). The `MapMessageToScalar` operation needs to be defined in a way that would guarantee consistency as much as possible and option B is IMO the best for that.

BasileiosKal:
Fix #245

Define input messages as octet strings and explicitly call `messages_to_scalars` in each operation.

Note this makes #266 obsolete, since it enforces consistency on the input arguments (everything are octets now).

Based on #267 to avoid the extra explicit checks (would be couple of steps)


BasileiosKal:
Discussed on the WG call on the 19th of June, will define the Input of `messages_to_scalars` to be typed elements instead of octet strings.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -586,8 +584,9 @@ Inputs:
                      it defaults to an empty string.
 - ph (OPTIONAL), an octet string containing the presentation header. If not
                  supplied, it defaults to an empty string.
-- disclosed_messages (OPTIONAL), a vector of scalars. If not supplied,
-                                 it defaults to the empty array "()".
+- disclosed_messages (OPTIONAL), a vector of input\_messages. If not
```
andrewwhitehead, line 587:
```suggestion
- disclosed_messages (OPTIONAL), a vector of input_messages. If not
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -741,37 +740,69 @@ Procedure:
...
 
-This operation takes an input message and maps it to a scalar value via a cryptographic hash function for the given curve. The operation takes also as an optional input a domain separation tag (dst). If a dst is not supplied, its value MUST default to the octet string returned from ciphersuite\_id || "MAP\_MSG\_TO\_SCALAR\_AS\_HASH\_", where ciphersuite\_id is the ASCII string representing the unique ID of the ciphersuite "MAP\_MSG\_TO\_SCALAR\_AS\_HASH\_" is an ASCII string comprised of 26 bytes.
+The signature of the operation is the following,
 
 ```
-msg_scalar = MapMessageToScalarAsHash(msg, dst)
+(scalar_1, ..., scalar_L) = messages_to_scalars((msg_1, ..., msg_L))
+```
+
+Every `messages_to_scalars` operation MUST define a unique `MESSAGES_2_SCALARS_ID` value to be used by the ciphersuite. This value MUST only contain ASCII encoded characters with codes between 0x21 and 0x7e (inclusive) and MUST end with an underscore (ASCII code: 0x5f), other than the last character the string MUST not contain any other underscores (ASCII code: 0x5f).
```
andrewwhitehead, line 755:
```suggestion
Every `messages_to_scalars` operation MUST define a unique `MESSAGES_TO_SCALARS_ID` value to be used by the ciphersuite. This value MUST only contain ASCII encoded characters with codes between 0x21 and 0x7e (inclusive) and MUST end with an underscore (ASCII code: 0x5f), other than the last character the string MUST not contain any other underscores (ASCII code: 0x5f).
```

I think it's better to spell it out.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -741,37 +740,69 @@ Procedure:
...
+2. for i in (1, ..., length(messages)):
+3.     msg_i = messages[i]
+3.     if msg_i is an octet string:
+4.         msg_scalar_i = hash_to_scalar(msg_i, dst)
+5.     else:
+6.         msg_scalar_i = msg_i
+7. return (msg_scalar_1, ..., msg_scalar_L)
+```
+
+The `MESSAGES_2_SCALARS_ID` of the `hash_messages_to_scalar` is defines as,
```
andrewwhitehead, line 792:
```suggestion
The `MESSAGES_TO_SCALARS_ID` of the `hash_messages_to_scalar` is defines as,
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1197,11 +1229,13 @@ This section defines the format for a BBS ciphersuite. It also gives concrete ci
 The following section defines the format of the unique identifier for the ciphersuite denoted `ciphersuite_id`, which will be represented as an ASCII encoded octet string. The REQUIRED format for this string is
 
 ```
-  "BBS_" || H2C_SUITE_ID || ADD_INFO
+  "BBS_" || H2C_SUITE_ID || MESSAGES_2_SCALARS_ID || ADD_INFO
```
andrewwhitehead, line 1232:
```suggestion
  "BBS_" || H2C_SUITE_ID || MESSAGES_TO_SCALARS_ID || ADD_INFO
```

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1197,11 +1229,13 @@ This section defines the format for a BBS ciphersuite. It also gives concrete ci
...
 ```
 
   *  H2C\_SUITE\_ID is the suite ID of the hash-to-curve suite used to define the hash_to_curve function.
 
+  *  MESSAGES\_2\_SCALARS\_ID is the ID of the messages\_to\_scalars operation, as defined in (#messages-to-scalars).
```
andrewwhitehead, line 1237:
```suggestion
  *  MESSAGES\_TO\_SCALARS\_ID is the ID of the messages\_to\_scalars operation, as defined in (#messages-to-scalars).
```

andrewwhitehead:
I put in some suggestions for using `MESSAGES_TO_SCALARS_ID`, but perhaps `OCTETS_TO_SCALARS_ID` would make more sense and avoid overuse of the term 'messages'?

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -136,15 +136,15 @@ R
 U
 : The number of message indexes that are undisclosed in a proof-of-knowledge of a signature.
 
-msg
-: An input message to be signed by the signature scheme.
+scalar
+: An integer between 0 and r-1, where r is the prime order of the selected groups, defined by each ciphersuite (see also [Notation](#notation)).
+
+input_message
+: An input message to be signed by the signature scheme. An input_message can either be either an octet string or a scalar.
```
andrewwhitehead, line 143:
```suggestion
input\_message
: An input message to be signed by the signature scheme. An input\_message can either be either an octet string or a scalar.
```

BasileiosKal:
Updated the operation. As discussed in the WG, added a `map_to_scalar` operation, that will map a `(msg, idx)` pair to a scalar.

1. Turns out that if a msg is a scalar, simply returning it (while hashing the octet strings) is not secure. There is no way to know that what was signed is the octet string `msg` and not the `scalar = hash_to_scalar(msg)`.
2. Did not use `octet_to_scalar` for the same reason.

cc @andrewwhitehead i think the PR is ready for re-review!

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -765,38 +764,104 @@ When defining a new `create_generators` procedure, the most important property i
...
+
+```
+MAP_TO_SCALAR_ID = "HM2S_"
+```
+
+### Define a new Map to Scalar
+
+To define different ways with which messages can be mapped to scalars, an application can define a new `map_to_scalar` operation, as part of a new ciphersuite. A new `map_to_scalar` function is REQUIRED to adhere to the following security rules:
+
+1. It MUST return unique values for different `msg` inputs.
```
andrewwhitehead, line 861: Within reasonable guarantees of collision avoidance?

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -765,38 +764,104 @@ When defining a new `create_generators` procedure, the most important property i
...
 
-### Message to Scalar as Hash
+#### Map to Scalar as Hash
 
-This operation takes an input message and maps it to a scalar value via a cryptographic hash function for the given curve. The operation takes also as an optional input a domain separation tag (dst). If a dst is not supplied, its value MUST default to the octet string returned from ciphersuite\_id || "MAP\_MSG\_TO\_SCALAR\_AS\_HASH\_", where ciphersuite\_id is the ASCII string representing the unique ID of the ciphersuite "MAP\_MSG\_TO\_SCALAR\_AS\_HASH\_" is an ASCII string comprised of 26 bytes.
+This document specifies the following `map_to_scalar` operation, called `map_to_scalar_as_hash`, that uses `hash_to_scalar` as defined in (#hash-to-scalar). Although for extendability reasons, the `map_to_scalar` operation accepts messages that can be either an octet string or a scalar value (as to support protocol specific preprocessing of a message), the `map_to_scalar_as_hash` operation used by this document only maps octet string to scalars and will abort if it gets an `input_message` that is already a scalar value. Additionally, the resulting scalar does not depend on the `index` of the message.
```
andrewwhitehead, line 818: Hm, this seems to mean that any application which wants to pre-process the messages as mentioned above, must use a ciphersuite not defined by this document, with a different mapping function.

BasileiosKal: Yeap that was the indent. It seems that there is no way to define protocol-agnostic handling of scalar values (i.e., something like `if msg a scalar, return msg` does not work). 

Rules need to be more strict and also consider the index of a message (for example the 1st message is expected to be a scalar, etc). So a ciphersuite defined protocol-specific mapping seems unavoidable.

BasileiosKal: You can always also define the `map_to_scalar` operation to just accept scalars and be the identity, and then apply whatever pre processing you want??

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Updated! Thanks!

christianpaquin:
As mentioned on the discussion in today's WG call: U-Prove allows both direct encoding and hashing the attributes. This is useful, within _one_ ciphersuite, to allow encoding attributes of different types in one credential (e.g.: hash name, address, etc., but encode directly age and expiration to allow range proofs). See the `e` values in the Issuer params (section 2.3.1 of the [crypto spec](https://github.com/microsoft/uprove-node-reference/blob/main/doc/U-Prove%20Cryptographic%20Specification%20V1.1%20Revision%205.pdf)).

BasileiosKal:
Discussed on the WG call on the 10th of July. Merging and will open an Issue to keep track of the conversation.

Issue #223, opened 2022-10-19
Tomislav's comments


BasileiosKal:
This is a tracking issue to address [comments](https://github.com/decentralized-identity/bbs-signature/pull/221#issuecomment-1283168657) made by @tmarkovski around test vectors and the spec in general!

tmarkovski:
Posting them here.

My thoughts and findings:

- Most negative tests are hard to validate for the reason they specify. Once my positive tests worked, I only assumed the negative tests are working as well, since they were all failing before.
- Tests without header and presentation header might be useful as well
I stumbled on the use of hex and treating messages as encoded scalars, but this has already been addressed I think. Might be useful to add variable length message inputs.
- Why does the Sign algorithm accept SK and PK as input, but doesn't validate if the PK is correct. SkToPk is fairly inexpensive operation, it seems that only SK is needed. Is there a security concern similar to what affected ECDSA?

tplooker:
> Most negative tests are hard to validate for the reason they specify. Once my positive tests worked, I only assumed the negative tests are working as well, since they were all failing before. The algorithm doesn't provide ways to test exactly where the failure happened.

Agreed, I think even if an implementation does not verify the specific reason for a failure when testing with the negative test vectors, they still provide some value. The "reason" for failure is really only there for informational purposes, so we should clarify that.

> Tests without header and presentation header might be useful as well

+1

> I stumbled on the use of hex and treating messages as encoded scalars, but this has already been addressed I think. Might be useful to add variable length message inputs.

Yes I believe this has been addressed 

> Why does the Sign algorithm accept SK and PK as input, but doesn't validate if the PK is correct. SkToPk is fairly inexpensive operation, it seems that only SK is needed. Is there a security concern similar to what affected ECDSA?

Correct it is because of performance reasons, not sure which security concern you are referring to with ECDSA, do you mean EdDSA. We have investigated that even without confirming the PK's relationship to the SK in sign there is no security concern, the signature would just fail to validate. Implementations are free to do this validation and we could add in an implementation consideration to that effect?

> Should PK parameters be passed as points to Verify, ProofGen, ProofVerify, instead as octet string? This will simplify the validation and be inline with how we pass the messages as scalars. Same question for SK.

I understand this perspective, however I think its good to be consistent, both proofs and signatures are passed in as their encoded form instead of as internal structures and I view PK the same.

tplooker:
Related to #7, we may want to merge with that so we dont have a duplicate.

[...]

PR #273, opened 2023-07-02
Update test vectors

BasileiosKal:
Test vectors for #267 and #268.

## Breaking changes from the PRs

1. From #267

    - `hash_to_scalar`:  removed `count`. There is no longer a loop and the input to `expand_message` is just `msg_octets` instead of `msg_prime = msg_octets || I2OSP(count, 8)`
    - `creat_generators`: removed `n` (and the check at step 8) and went from `I2OSP(n, 4)` to `I2OSP(i, 8)` in the input to `expand_message` when calculating `v`.

2. From #268

    - `ciphersuite_id`: added `HM2S_` at the end (the `MAP_TO_SCALAR_ID` of the `map_to_scalar` operation).


BasileiosKal:
Note that the test vectors use variable length messages now (with one of the messages being empty). Addresses one of the comments in #223.

BasileiosKal:
Discussed on the WG call of the 10th of July. Fixtures are cross validated. Mergin

PR #277, opened 2023-07-10
Update publish workflow

BasileiosKal:
Fix (hopefully) the test vectors not appearing on the [published draft](https://www.ietf.org/archive/id/draft-irtf-cfrg-bbs-signatures-02.html) issue.


BasileiosKal:
Discussed on the WG call on the 10th of July. Consensus on merging it. Merging

PR #290, opened 2023-10-23
Clarify zero indexing of arrays

BasileiosKal:
Not a breaking change. Implementations used 0-indexing already,. even though was not clearly specified in the draft.


BasileiosKal:
Discussed on the WG call on the 23rd of Oct. Will resolve conflicts and merge after approval.

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -176,12 +176,12 @@ a || b
 I \\ J
 : For sets I and J, denotes the difference of the two sets i.e., all the elements of I that do not appear in J, in the same order as they were in I.
 
+X\[i\]
+: Denotes the element of array `X` at index `i`. Note that arrays in this document are considered "zero-indexed", meaning that element indexing starts from 0 rather than 1. For example, if `X = [a, b, c, d]` then `X[0] = a`, `X[1] = b`, `X[2] = c` and `X[3] = d`.
```
dwaite, line 180: I would propose this note be moved down to the middle of the notation section (before mathematical notation), and be expanded with specific examples between the algorithmic and mathematical notations.

Proposed text:

~~~
Note that arrays in this document use the zero-based numbering common in
many programming languages, meaning that element indexing starts from 0.
This is distinct from naming used for explaining mathematical concepts,
where natural (one-based) numbering might be used as part of the names
for clarity in that context.

For example,

```
X = [a_1, a_2, ..., a_n]
```

would indicate

```
X[0] = a_1
X[1] = a_2
// ... and so on, up to
X[n-1] = a_n
```
~~~

dwaite (changes requested):

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Thanks for the suggestion 🙏 

For now, added the suggested text to the `Considerations` section, to keep the notation text short.

Wind4Greg (approved):

andrewwhitehead (approved):

PR #291, opened 2023-10-29
Add privacy considerations

### LINKED ISSUE #256: Add a privacy consideration around revealed message indicies
Each message that is signed when producing a signature is associated to a specific generator. This relationship has to persist for revealed messages in any derived proofs, which means the following pieces of information are known to a verifier when verifying a proof.

1) The total number of messages that were originally signed (even if only a subset are actually revealed).
2) What index each of the revealed messages in a generated proof occupied in the originally signed set.
3) By process of elimination the indicies for all the messages that were not revealed in a derived proof.

We should add a privacy consideration that describes this so that implementations can be aware in case it leads to revealing information unintentionally to a verifier.

BasileiosKal:
Closes #256.


draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1578,6 +1578,26 @@ Procedure:
...
...ges list. Second, the index the disclosed messages had in the list of signed messages (see (#proof-generation-proofgen)). This information, if unique to each Prover, could be employed to correlate multiple proof presentations together. As a result, the Issuer should not sign lists of messages with unique lengths or unique indexing. For this reason, it is RECOMMENDED that signed lists of messages are padded to a common length (using either random, or an unused by the application message, like 0 or 1). It is also RECOMMENDED that a constant ordering of messages will be preserved when possible. For example, if an application creates signatures for the messages `[<user_name>, <user_affiliation>, <user_country>]`, then those messages should always be signed in the same order, i.e., first message should always be the user's name (`<user_name>`), second message should always be the user's affiliation (`<user_affiliation>`) and the last message should always be the user's country of origins (`<user_county>`). Provers can employ consistency validation mechanisms, like the ones described in [@I-D.ietf-privacypass-key-consistency], to validate that those values are not used to correlate them.
```
dwaite, line 1589:
```suggestion
When a Prover presents a BBS proof to a Verifier, other than the messages they decide to disclose, there are two additional pieces of information that will be revealed. First, the total number of signed messages, which can be inferred from the size of the BBS proof and the length of the disclosed messages list. Second, the index the disclosed messages had in the list of signed messages (see (#proof-generation-proofgen)). This information, if unique to each Prover, could be employed to correlate multiple proof presentations together. As a result, the Issuer should not sign lists of messages with unique lengths or unique indexing. For this reason, it is RECOMMENDED that signed lists of messages are padded to a common length (using either random, or an unused by the application message, like 0 or 1). It is also RECOMMENDED that a constant ordering of messages will be preserved when possible. For example, if an application creates signatures for the messages `[<user_name>, <user_affiliation>, <user_country>]`, then those messages should always be signed in the same order, i.e., first message should always be the user's name (`<user_name>`), second message should always be the user's affiliation (`<user_affiliation>`) and the last message should always be the user's country of origins (`<user_country>`). Provers can employ consistency validation mechanisms, like the ones described in [@I-D.ietf-privacypass-key-consistency], to validate that those values are not used to correlate them.
```

dwaite: user_county -> user_country

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1578,6 +1578,26 @@ Procedure:
 5. return W
 ```
 
+# Privacy Considerations
+
+This section will go through threats to the Prover's privacy. Note that a BBS proof is unlinkable both against Verifiers coalition and Verifiers coalition with the Issuer. The following sections will describe possible threats, resulting from side chanel information or identifying disclosed messages, that could compromise the unlinkability property of the BBS proof. Such threats, if exploited, could lead to correlation of the Prover's interactions with different Verifiers, resulting to fingerprinting attacks on the Prover's activity.
```
dwaite, line 1583: Is the typical term in this context coalition or correlation?

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Thanks! Fixed.

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: AFAIK yes. Correlation generally refers to breaking the unlinkability property of the proofs (i.e., "correlating" multiple proofs presenting by the same Prover together)

Coalition refers to multiple parties (Verifiers and potentially the Issuer) working together to track the user's activity (the proofs they produce).

Wind4Greg (approved):

BasileiosKal:
Discussed on the WG 20th of Nov. Will update wording to use "collusion" instead of "coalition".

andrewwhitehead (approved):

BasileiosKal:
Discussed on the WG at the 4th of December. Multiple approvals. Editorial updates. Merging! Thank you!

PR #296, opened 2023-11-27
Update ciphersuites

BasileiosKal:
- Removes the hash variable from the ciphersuites.
- Fixes the ciphersuite IDs


BasileiosKal:
Multiple approvals. Editorial update. Merging.

PR #298, opened 2023-11-28
Various editorial updates

BasileiosKal:
Couple of editorial updates
- Updated abstract and some security considerations
- Added academic works overview
- Added header and presentation header usage section
- Added section on post-quantum security


draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1668,11 +1660,17 @@ Applications using the Interface defined in (#bbs-signatures-interface), MUST ig
...
+
+## Post-quantum Security and Everlasting Privacy
+
+The BBS Signatures scheme has two main security properties; unforgeability of BBS signatures and zero-knowledge of BBS proofs. The first property, referring to the inability of anyone other than the Signer to be able to generate BBS signatures, does not hold true against an adversary using a strong enough quantum computer. Specifically, such a computer could be used to reveal the Signer's secret key from their public key, hence giving them the ability to generate BBS signatures on behalf of that Signer for any message they want.
+
+On the other hand, the second security property, i.e., the zero-knowledge of the BBS proof, cannot be broken, even by utilizing a quantum computer. We refer to this property as "everlasting privacy". This guarantees that an adversary in possession of a quantum computer, will not be able to reveal neither the messages undisclosed by a BBS proofs, nor the hidden signature value. As a result, the privacy and hiding properties of the BBS proofs that are currently used, will not be compromised by future quantum-attacks.
```
OR13, line 1673: you probably want to use terminology from PQUIP here.

https://datatracker.ietf.org/doc/draft-ar-pquip-pqc-engineers/

CRQCs, etc.

OR13 (approved):

draft-irtf-cfrg-bbs-signatures.md
BasileiosKal: Than you for the suggestion! Updated.

andrewwhitehead (approved):

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -50,25 +50,23 @@ organization = "CryptID"
...
+This document describes the BBS Signature scheme, a secure, multi-message digital signature protocol, supporting proving knowledge of a signature while selectively disclosing any subset of the signed messages. Concretely, the scheme allows for signing multiple messages whilst producing a single, constant size, digital signature. Additionally, the possessor of a BBS signatures is able to create zero-knowledge, proofs-of-knowledge of a signature, while selectively disclosing subsets of the signed messages. Been zero-knowledge, the BBS proofs do not reveal any information about the undisclosed messages or the signature it self, while at the same time, guarantying the authenticity and integrity of the disclosed messages.
```
Wind4Greg, line 53: This document describes the BBS Signature scheme, a secure, multi-message digital signature protocol, supporting proving knowledge of a signature while selectively disclosing any subset of the signed messages. Concretely, the scheme allows for signing multiple messages whilst producing a single, constant size, digital signature. Additionally, the possessor of a BBS signatures is able to create zero-knowledge, proofs-of-knowledge of a signature, while selectively disclosing subsets of the signed messages. Being zero-knowledge, the BBS proofs do not reveal any information about the undisclosed messages or the signature it self, while at the same time, guarantying the authenticity and integrity of the disclosed messages.

Wind4Greg: Was just trying to correct the word usage: "Been" ==> "Being".

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -315,6 +317,12 @@ We also use strings in double quotes to represent ASCII-encoded literals. For ex
 
 Those rules will be used explicitly on every operation. See also [Serialize](#serialize).
 
+### Header and Presentation Header Usage
+
+There are two special values defined by the BBS Scheme; the `header` and the `presentation_header`. The `header` value is chosen by the Signer and is bound to both a BBS signature and the BBS proofs generated using that signature. Specifically, the Prover is required to reveal the `header` to the proof Verifier, during every BBS proof presentation. As a result, the Signer SHOULD NOT include to the `header` any identifying information, that may have the potential of compromising the Prover's privacy. Suitable use cases taking advantage of the `header` value, are binding a BBS signature (and subsequent BBS proofs) to a specific application, deployment or domain, or binding the signature to specific sets of metadata etc..
```
Wind4Greg, line 322: There are two special values defined by the BBS Scheme; the `header` and the `presentation_header`. The `header` value is chosen by the Signer and is bound to both a BBS signature and the BBS proofs generated using that signature. Specifically, the Prover is required to reveal the `header` to the proof Verifier, during every BBS proof presentation. As a result, the Signer SHOULD NOT include in the `header` any identifying information, that may have the potential of compromising the Prover's privacy. Suitable use cases taking advantage of the `header` value, are binding a BBS signature (and subsequent BBS proofs) to a specific application, deployment or domain, or binding the signature to specific sets of metadata etc..

Wind4Greg: Change:
"the Signer SHOULD NOT include to the `header` any" ==> "the Signer SHOULD NOT include in the `header` any"

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -315,6 +317,12 @@ We also use strings in double quotes to represent ASCII-encoded literals. For ex
...
+
+Similarly, the Prover can choose a `presentation_header` value to be bound to the BBS proof (in contrast to the `header` value that is chosen by the Signer and is bound to both BBS proof and signature). Verifying a BBS proof will guarantee the authenticity and integrity of the `presentation_header`. This makes it suitable for applications where the Prover may want to "sign" a message in addition to creating their BBS proof, in a way similar to how group signatures may work. Specifically, the Verifier will know that the message included in the `presentation_header` was signed by a member of the group, where group here the Provers having received valid signatures from the specific Signer, but will not know from exactly which member of the group the message was signed. Another application of the `presentation_header` is to guarantee freshness of a BBS proof and to avoid replay attacks. An example of how this could be achieved, is by including a random value supplied by the Verifier to the `presentation_header` value.
```
Wind4Greg, line 324: Similarly, the Prover can choose a `presentation_header` value to be bound to the BBS proof (in contrast to the `header` value that is chosen by the Signer and is bound to both BBS proof and signature). Verifying a BBS proof will guarantee the authenticity and integrity of the `presentation_header`. This makes it suitable for applications where the Prover may want to "sign" a message in addition to creating their BBS proof, in a way similar to how group signatures can work. Specifically, the Verifier will know that the message included in the `presentation_header` was "signed" by a member of the group, where group here being the Provers having received valid signatures from the specific Signer, but will not know from exactly which member of the group the message was signed. Another application of the `presentation_header` is to guarantee freshness of a BBS proof and to avoid replay attacks. An example of how this could be achieved, is by including a random value supplied by the Verifier to the `presentation_header` value.

Wind4Greg: Just some small English fixes in the above.

Wind4Greg (approved):
Tried to put in suggest some minor wording/English fixes. If they don't come through we can catch them during the "nits" phase.

BasileiosKal:
Multiple approvals. Addressed review comments. Discussed on the WG call at the 11th of Dec. Merging. Thank you all!

PR #305, opened 2023-12-13
Making Pairing Friendly Curves draft reference informative

BasileiosKal:
Redefine the relative functionality from [Pairing-Friendly Curves](https://datatracker.ietf.org/doc/draft-irtf-cfrg-pairing-friendly-curves/) in this draft.


OR13 (approved):
even an informative reference to an expired ID will probably cause problems.

BasileiosKal:
> even an informative reference to an expired ID will probably cause problems.

That's true, but it is not "forbidden" (see the [bls sigs](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05) reference to the [h2c rfc](https://www.rfc-editor.org/rfc/rfc9380.html#I-D.irtf-cfrg-bls-signature)). Even a normative reference is not an issue for the crypto panel review phase (it will certainly be a problem in later stages though).

Wind4Greg (approved):

andrewwhitehead (approved):

BasileiosKal:
Discussed on the WG call at the 18th of Dec. Will resolve conflicts and merge after that.

BasileiosKal:
Conflicts resolved. Multiple approvals. Merging. Thank you all!

PR #319, opened 2024-06-03
Bound memory improvements

### LINKED ISSUE #313: Support Bounded-Memory Implementations with a single Message Pass
Hi all,

I am trying to implement the IETF-draft in C using no dynamically allocated memory, thus suitable for e.g. embedded devices with no heap implementation.

There are only a few places in the specification which are troublesome to such an endeavour and I wanted to share these with you, and suggest improvements, should you wish to explicitly support this use case.

For efficiency, there are two critical points to consider:

Ideally, it should be possible to only go over every message once per high level API call (Sign, Verify, ProofGen, ProofVerify). Among other things, this requires being able to generate message scalars individually, given a message's index and optionally the total number of messages (which the current draft supports).

Likewise, it would be helpful if one could avoid generating the generators more than once during such API calls.

Let us briefly go over every high-level API function dealing with messages to see how this would work. Below algorithm sketches rely on the ability to incrementally hash arbitrary octet strings, but this is typical with many hashing APIs.

### Sign

`Sign` can be implemented with a single message pass. While iterating over the messages, the algorithm would use `B` as an accumulator, and incrementally prepare the input to hash into `e` (as part of `hash_to_scalar`).

`Sign` does not allow to only generate generators once when doing this: Before adding any messages to be hashed into `e`, the `domain` must be added, for which all generators must already be generated. Then while accumulating `B`, the generators need to be regenerated, as there is no way to store an arbitrary amount of generators in bounded memory.

A simple fix would be to reorder the inputs to the generation of `e`:

```
2. e = hash_to_scalar(serialize((SK, msg_1, ..., msg_L, domain)), signature_dst)
```

This should not have an effect on security, as `e` is essentially only supposed to be "random" but unique for each message. But please double-check :)

### Verify

`Verify` can be implemented efficiently due to commutativity of field elements. Essentially, one iterates over the messages, generates generators on the fly, accumulates "generator * message_scalar" into `B` while adding the generators to the domain-Hash, and finally adds `Q_1 * domain` to `B`.

### ProofGen

Other than some ideas from above, this function runs into two issues, one of which is easily fixed.

1. The `m_tilde` values need to be generated before the the challenge phase (accumulated into `T2` and hashed into the challenge) and after it (to generate `m_hat`s, which are part of the signature). Fortunately, the values are essentially pseudorandom, and can deterministically be regenerated from a PRG at little to no cost. An algorithm would store the undisclosed message scalars temporarily into the proof, and after the challenge phase multiply it with the challenge and add `m_tilde`.

2. The challenge calculation hashes first `Bbar` and friends, then the messages. Because the messages are required for the calculation of `B`, an implementation has to iterate over them multiple times. The fix involves a reordering of the inputs to the random oracle, which again should not have an impact on security. The ideal ordering would look like this:

```
1. c_arr = (R, i1, msg_i1, ..., iR, msg_iR, Abar, Bbar, D, T1, T2, domain)
```

### ProofVerify

Besides the problems already mentioned, this function seems fine. In particular, the modified challenge calculation should work here as well.

## TL;DR

By reording a few inputs to hash functions in two places, low-resource implementations can become potentially a lot more efficient. Thanks for reading :)

BasileiosKal:
Fixes #313

\+ new test vectors


Wind4Greg:
Great! Can you brief us at today's meeting?

draft-irtf-cfrg-bbs-signatures.md
```diff
@@ -1110,8 +1110,8 @@ ABORT if:
 
 Procedure:
 
-1. c_arr = (Abar, Bbar, D, T1, T2, R, i1, ..., iR,
-                                            msg_i1, ..., msg_iR, domain)
+1. c_arr = (R, i1, msg_i1, ..., iR, msg_iR, domain,
```
Wind4Greg, line 1113: 1. c_arr = (R, i1, ..., iR, msg_i1, ..., msg_iR, domain,

Wind4Greg: My mistake. Didn't understand the syntax.

BasileiosKal: Added an extended description of the inputs to the challenge digest calculation. Hopefully this should better explain the intended ordering of the index, message pair in `c_arr`.

Wind4Greg (approved):
Nice clear, higher level description of the challenge computation.

andrewwhitehead (approved):

BasileiosKal:
Multiple approvals. Merging! Thank you all!

PR #320, opened 2024-06-25
Editorial updates for draft 06

### LINKED ISSUE #317: DST collisions
There is a collision in the domain separation tags in CoreSign (used in the calculation of `e`):

```
1. signature_dst, an octet string representing the domain separation
                  tag: api_id || "H2S_" where "H2S_" is an ASCII string
                  comprised of 4 bytes.
```

And in ProofChallengeCalculate:

```
1. challenge_dst, an octet string representing the domain separation
                  tag: api_id || "H2S_" where "H2S_" is an ASCII string
                  comprised of 4 bytes.
```

And in calculate_domain:

```
1. domain_dst, an octet string representing the domain separation tag:
               api_id || "H2S_" where "H2S_" is an ASCII string
               comprised of 4 bytes.
```

Is there any particular reason for these collisions?

I believe the corresponding `hash_to_scalar` invocations represent independent random oracles. If one really wants collisions to save on DSTs, the only place where it seems rather harmless is to make domain_dst equal to map_dst used in messages_to_scalars, since the inputs to calculate_domain then become just another message. But even then, adding additional DSTs does not seem too expensive.

One could also argue that the octets representing input to these functions never collide, but this argument seems weaker than to have different DSTs for each of them.

BasileiosKal:
Also fixes #317

### DIFF
--- a/draft-irtf-cfrg-bbs-signatures.md
+++ b/draft-irtf-cfrg-bbs-signatures.md
@@ -198,7 +198,7 @@ G1, G2
 GT
 : a subgroup, of prime order r, of the multiplicative group of a field extension.
 
-e
+h
 : G1 x G2 -> GT: a non-degenerate bilinear map.
 
 r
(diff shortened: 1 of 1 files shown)


Wind4Greg (approved):
Looks good Vasilis!

tplooker (approved):

BasileiosKal:
Multiple approvals! Merging. Thank you all 🙏

PR #326, opened 2024-09-12
Editorial review

Wind4Greg:
Some small edits for readability. Started from end of document.

1. Should use `NIZK` for acronym rather than `nizk`
2. Removed "subject to change" warning on additional test vectors.
3. Clarified scalar multiplication in *Notation* section. Document is very consistent with always using P*x where P is group element and x is scalar, in that order. *Notation* section had this reversed.


BasileiosKal (approved):

tplooker (approved):

BasileiosKal:
Editorial updates. Multiple approvals. Merging. Thank you all 🙏

